PR #1, opened 2012-06-05
For legacy_root, ensure that template class is added to #x-root

vmarta:
Executing `Mobify.$('html, #x-root').last()` from within the `<head>` would always give you the html element because #x-root has not existed yet.


jansepar:
+2 @romanrudenko and I

PR #4, opened 2012-06-07
Correctly reset the mobify-path cookie for Mobify.desktop()

vmarta:
Calling `Mobify.desktop()` didn't reset `mobify-path` cookie.  Thus, you kept seeing the mobile view.

@tedtate has fixed this a month ago for 1.0 tag but the update wasn't brought in to 1.1.


jansepar:
+2 from @romanrudenko and I

PR #6, opened 2012-06-07
Removing templateName from html element in legacyRoot

jansepar:
This merges that went into 1.1 should also go into 1.2. Leaving for @romanrudenko to review, as these merges are not urgent.


rrjamie:
This is already in back-to-desktop, closing.

PR #16, opened 2012-06-15
Combo

noahadams:
Integrates use of combo with localStorage based caching, and some tests of said feature


johnboxall:
`$.fn.combineScripts` should probably handle the cases where: No scripts were found / No external scripts were found.

Also, let's say `localStorage` isn't available - are we looking at a performance penalty? If so, if persistence isn't available, maybe we should just return the scripts we find in order.

api/combo-caching.js
```diff
@@ -164,12 +175,15 @@ var ccPublic = /^\s*public\s*$/
                 }
             }
 
+            //DEBUG
+            // console.log('urls: ' + JSON.stringify(urls));
+            // console.log('notCachedUrls: ' + JSON.stringify(notCachedUrls));
+
```
rrjamie: Remove // Debug comments if possible.

rrjamie:
What's the status of this?

johnboxall:
@rrjamie the latest code is awaiting review to be merged.

api/combo.js
```diff
@@ -0,0 +1,330 @@
...
+                    (!cacheControl['no-cache'])) {
+                    // Convert the max-age directive to ms.
+                    return now > (date + (cacheControl['max-age'] * 1000));
+                }
+            }
+
+            // If `expires` is present, we are stale if we are older.
+            if (expires) {
+                return now > Date.parse(expires);
+            }
```
rrjamie, line 206: Is this what we want? For non-parsables to never be stale:

```
 >>> 1 > Date.parse("Barf Barf Barf")
 false
```

api/combo.js
```diff
@@ -0,0 +1,330 @@
...
+
+            // Otherwise, we are stale.
+            return false;
```
rrjamie: If we are stale, return true?

PR #18, opened 2012-06-19
New documentation in place for review.

mezzoblue:
Have re-written the technical documentation to be more user-friendly. Would appreciate a second and third pair of eyeballs on this before merging, lots of stuff changed.

Most significant things to note:
- Added new 'Using Konf' section to konf reference, could be considered a candidate for replacing the entirety of 'Understanding Konf'
- fleshed out Reserved Keys in konf reference
- Added glossary, appendix
- Standardized wording in templates (variables and attributes)
- dropped sections in templates, seemed redundant when we already have variables and attributes
- added tons of code examples

Probably a bunch more I'm missing too.

### DIFF
--- a/docs/appendix.md
+++ b/docs/appendix.md
@@ -7,79 +7,70 @@
 2. Common Errors
 3. Dust.js Documentation
 
-
-1. Debugging
+# Debugging
 
 ## Tools
 
-Debugging Mobify.js adaptations requires the use of an advanced web inspection tool like Webkit Inspector or Firebug. You're likely well-acquainted with one of these, but here are a few quick start tips if not.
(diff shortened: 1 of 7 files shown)


jbandelin:
In the Glossary doc, the definition of context is more verbose than any of the other definitions. 

Do we need anything more than the last sentence "A view context is a variable name/value mapping that is passed on to the template" for this definitions?

jbandelin:
There is a typo in the appendix file. Under Debugging the Konf, in the second paragraph:

 "Look for a the item 'All Extracted Data', which..."

rrjamie:
I assume we are all pro-semicolon? If so, we should probably make sure we have semicolons throughout the code examples.

PR #32, opened 2012-07-12
Expose project name

noahadams:
Changes to mobifyjs API  and build necessary to expose the project name on imageresize's defaults object, note, requires modifications to mobify client to build a mobify.js that does this (these modifications have been completed on a similarly named branch of mobify-client)

### DIFF
--- a/api/resizeImages.js
+++ b/api/resizeImages.js
@@ -15,8 +15,6 @@ var absolutify = document.createElement('a')
       , '//ir3.mobify.com'
     ]
 
-    , projectName = Mobify.conf.projectName || ""
-
     /**
      * Hash `url` into a well distributed int.
      */
--- a/api/resizeImages.js
+++ b/api/resizeImages.js
@@ -46,8 +46,8 @@ var absolutify = document.createElement('a')
           , bits = [host];
 
         // If projectName is set on defaults and truthy, put it in resized image urls
-        if(defaults.projectName) {
-            var projectIdeintifier = "project-" + defaults.projectName
+        if (defaults.projectName) {
+            var projectIdeintifier = "project-" + defaults.projectName;
             bits.push(projectIdeintifier);
         }
 
@@ -94,9 +94,10 @@ var absolutify = document.createElement('a')
         return $imgs.each(function() {
             if (attr = this.getAttribute(opts.attribute)) {
                 absolutify.href = attr;
+                var url = absolutify.href;
                 // Produce an image resize url only for matched protocols
-                if(protocolMatcher.exec(absolutify.href)) {
-                    this.setAttribute('x-src', getImageURL(absolutify.href, opts))
+                if(protocolMatcher.exec(url)) {
+                    this.setAttribute('x-src', getImageURL(url, opts));
                 }
             }
         });
--- a/api/resizeImages.js
+++ b/api/resizeImages.js
@@ -46,7 +46,7 @@ var absolutify = document.createElement('a')
           , bits = [host];
 
         // If a projectName is set on the conf object, put it in resized image urls
-        if(Mobify.conf && && Mobify.conf.data && Mobify.conf.data.projectName) {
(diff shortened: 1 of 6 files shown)


noahadams:
Also includes a change to imageresize so it will only rewrite http(s) urls

api/resizeImages.js
```diff
@@ -43,6 +45,12 @@ var absolutify = document.createElement('a')
         var host = hosts[URLHash(url) % hosts.length]
           , bits = [host];
 
+        // If projectName is set on defaults and truthy, put it in resized image urls
+        if(defaults.projectName) {
```
jansepar: Space after if :P

konf/api_config.konf
```diff
@@ -0,0 +1,15 @@
...
+{?project_name}
+
+var projectName = '{project_name}';
+
+{! Add it to image resizing defaults !}
+Mobify.$.fn.resizeImages.defaults.projectName = projectName;
+
+{/project_name}
+
+})(window.Mobify);
```
jansepar, line 15: Was there a reason this has it's own file? Couldn't we just add it to config.js? Mobify.config is an accessible key that can be used as needed. Also why is it a .konf file living in the konf folder? I would say it should be a .js file living inside of the api folder.

noahadams, line 15: Because it's not executable js on its own, it's a place to macro in configuration variables. It's as much a "konf" file as base_konf.konf is.

I was initially abusing default.konf for this, but that's the wrong place for this to live, as all you can do there is add to defaults on the Mobify.conf.data object for your adaptation, whereas this file holds compile-time configuration parameters used by functions in the mobifyjs API themselves.

config.js does run-time configuration.

api/resizeImages.js
```diff
@@ -86,13 +94,17 @@ var absolutify = document.createElement('a')
         return $imgs.each(function() {
             if (attr = this.getAttribute(opts.attribute)) {
                 absolutify.href = attr;
-                this.setAttribute('x-src', getImageURL(absolutify.href, opts))
+                // Produce an image resize url only for matched protocols
+                if(protocolMatcher.exec(absolutify.href)) {
+                    this.setAttribute('x-src', getImageURL(absolutify.href, opts))
```
rrjamie: It might be a performance hit to touch `absolutify` twice, maybe save it in to a var.

tests/index.html
```diff
@@ -381,12 +387,25 @@ <h2 id="qunit-userAgent"></h2>
...
 
             equal($got.length, 1);
         });
 
+        test('$.fn.resizeImages non-http-url', function() {
+            expect(1);
+
+            $('#resizeImages-non-http').resizeImages()
+            var i = $('#resizeImages-non-http img')
+            var url = i.attr('x-src')
```
rrjamie: semi-semi-semi colons.

konf/api_config.konf
jansepar, line 15: Hrmmm. So, I guess it makes sense that it's not in config.js because config.js is used for run-time configuration, but I really don't think it makes sense as being a konf file. It follows the same (function(Mobify){})() pattern as the rest of the api files. I would argue that it is either a js file that lives inside of api/, or it is just something we should add to default.konf. 

I dont see what's wrong with adding that data into the Mobify.conf object, can't you just grab that variable when you need it inside of the resizeImage function? Default.konf already contains compile-time configuration variables like {build_dt}.

noahadams, line 15: Logically it should live in something like defaults.konf, but the way defaults.konf is templated and evaluated (as part of konf evaluation) makes it weird to  put API config in it. 

It makes the API bits substantially less self-contained as they need to peek in to Mobify.conf.data at conf evaluation time to get their parameters, as opposed to this, where the parameter is exposed and set imperatively here.

I've only named it ".konf" because that's the file extension we use for dusty javascript.

jansepar, line 15: Alright, I can buy that :)

PR #41, opened 2012-07-27
Updates the `context.choose` reference.

johnboxall:
Also adds a doc for "Getting Started" with the Cloud.

### DIFF
--- a/www/docs/cloud-getting-started.md
+++ b/www/docs/cloud-getting-started.md
@@ -1,0 +1,224 @@
+---
+layout: doc
+title: Getting Started with the Mobify Cloud
+---
+
+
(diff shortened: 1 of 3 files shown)


rrjamie:
+1.

jansepar:
I think we need to simplify the way choose is explained. I think the in-depth description can be left on the konf reference (and maybe in the getting started there is a link to "learn choose more in-depth"). I have found that the easiest way to explain choose to people is to say "You have a list of template objects that describe which template to use, and what must match in the DOM for it to render.".

How about we just merge this pull request, and I will update it to how I want it to be described and issue a separate pull request.

PR #50, opened 2012-08-02
Bad style merge made the slides ugly. Fixing.

mezzoblue:


rrjamie:
If we are going for batteries-included styles here, we should be probably be quite conservative with the number of different classes we provide.

m-card, m-card-light, m-slide, etc.

mezzoblue:
I'm not sure where m-card-light came from in the first place, but it looks like the ugliness this branch is fixing was introduced when someone tried to merge the m-slide and m-card styles. Since m-slide basically replaced m-card-light anyway, I'm killing it.

rrjamie:
+1.

PR #51, opened 2012-08-02
Accordion update

mezzoblue:
Adjusting Accordion examples a bit to fix some errors, streamline them, and added a new arrow example at the bottom


rrjamie:
Do we expect any confusion because the colours are inline styes? Eg, if I copy and paste the markup for an the redish example, will I be disappointed with my sad grey accordion?

Possible solution: remove the colours.

mezzoblue:
I see it as more of a demonstration of how you extend the base style. Here's a module, it looks like this. Want to make it look different? We did in this example, you can study our code to see how it was done.

jansepar:
+1 looks good with the z-index stuff hiding the arrow upon closing :D

PR #52, opened 2012-08-02
Zip the correct carousel files for download

vmarta:
It was including accordion stuffs for some reason. And it had the
unnecessary(?) carousel-controls.css.


jansepar:
Doing a bunch of changes to how the zips will be managed. They will no longer be committed. Thanks for the pull request though! Closing.

PR #56, opened 2012-08-03
The module reorganization borked the accordion example

mezzoblue:


jansepar:
This does fix the problem, but the problem has already been fixed and merged into master :). Closing!

Issue #61, opened 2012-08-13
Add image reszing doc to mobifyjs repo


noahadams:
With cloud feature disclaimer added.

gane5h:
PR 64 was accepted and merged back into master.

Issue #62, opened 2012-08-13
Add combo doc to mobifyjs docs


noahadams:
With cloud feature disclaimer.

gane5h:
PR 64 was accepted and merged back into master.

PR #64, opened 2012-08-15
Imageresize combo docs

noahadams:
Would appreciate some feedback on the addition of these documents, the copy and perhaps some styling on the callout about the documents being for cloud features.


noahadams:
will resolve #61, #62

gane5h:
Can't I use the CSS namespacing features through Mobify.js?

noahadams:
There's no API exposed for it in MobifyJS.

PR #65, opened 2012-08-15
Imageresize no sharding

### LINKED ISSUE #57: Make not using shards the default in $.fn.resizeImages() url rewriting
Domain sharding doesn't appear to provide any performance gains in modern mobile browsers, as they will open more simultaneous connections to a single domain.

noahadams:
Removes sharding from image resize client code and updates tests to reflect that.

Resolves #57

### DIFF
--- a/api/resizeImages.js
+++ b/api/resizeImages.js
@@ -1,25 +1,29 @@
-(function(window, $) {
+/**
+ * Mobify.js API to the Mobify Image Resizing Service.
+ */
+(function(window, Mobify, Math) {
 
-var absolutify = document.createElement('a')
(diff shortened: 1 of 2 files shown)


api/resizeImages.js
```diff
@@ -77,8 +48,9 @@ var absolutify = document.createElement('a')
           , $imgs = this.filter(opts.selector).add(this.find(opts.selector))
           , attr;
 
+        // integer width
         if (typeof options == 'number') {
-            opts.maxWidth = options;
+            opts.maxWidth = Math.floor(options);
```
johnboxall, line 57: @noahadams why are we flooring this arg?

PR #77, opened 2012-09-24
add url patterns into mobifyjs

noahadams:
Allows studio generated projects to be successfully previewed/built with mobify-client. Manually verified working.


johnboxall:
Thanks for the feed back @mlowicki.

This request would have added compatibility with code generated from cloud.mobify.com. We've decided to hold off on committing it to Mobify.js until it is more inline with the existing standards of this project!
- Written in JavaScript
- Human friendly API
- Concise functionality for URL based template routing

PR #80, opened 2012-10-02
Adding a "Back to Mobile" section in the docs.

jansepar:


noahadams:
+1 Markdown compiles properly, also, suggested technique tested and working.

PR #83, opened 2012-11-26
Tweak the behaviour of `$.fn.resizeImages`.

johnboxall:
- Expose the resize host to allow it to be changed for testing.
- Return only images that were modified.


noahadams:
Looks good, tests pass, :+1:

PR #87, opened 2013-01-04
Add optional image cache expiry to the imageResize API

b1tr0t:


b1tr0t:
@jansepar Can you have a look at this for me?

api/resizeImages.js
```diff
@@ -26,6 +26,10 @@ var $ = Mobify.$
             bits.push(projectId);
         }
 
+        if (options.cacheHours && typeof(options.cacheHours) === 'number') {
```
jansepar: If someone sets options.cacheHours as a string, then this will silently fail. Probably bad because I would imagine some people might set the value as a string, for example:

`options.cacheHours = 27;`
`options.cacheHours = '27';`

I feel like both of those should work. If the argument IS a string, we can check with `isNaN` or `parseInt` to determine if it is a valid number, and throw an exception if it is not a number.

b1tr0t:
@jansepar Integrated your feedback

johnboxall:
@b1tr0t we can rely on the backend to provide error reporting and for the most part, JavaScript's type conversion will do the right thing here.

PR #88, opened 2013-01-08
Cleanup www.mobifyjs.com

johnboxall:
- Add <meta> attributes
- Remove unused JavaScript
- Delete unused files
- Update video links to the correct place


www/docs/index.md
```diff
@@ -38,8 +42,8 @@ Insert the Mobify.js tag **immediately** after the opening _<head>_ tag on the w
     document.write('<plaintext style="display:none">');
 
     setTimeout(function() {
-        var mobifyjs = document.createElement('script')
-          , script = document.getElementsByTagName('script')[0];
+        var mobifyjs = document.createElement('script');
+        var script = document.getElementsByTagName('script')[0];
```
jansepar, line 46: Are we changing this convention?

johnboxall, line 46: Yup.

jansepar, line 46: Why?

johnboxall, line 46: I find this format, though more verbose, lets you make less decisions per line. Declaring a variable? Use `var x;`. Always. This way you never have to think about whether you should put a semicolon on the end of the line, how you should indent, whether you should use a `var` etc...

jansepar, line 46: Agreed. I'm happy with this change in convention.

PR #89, opened 2013-01-14
Sticky reset for 5 minutes

b1tr0t:
When reset fires we should make this setting sticky for a few minutes to avoid the extra overhead of resetting when the site isn't yet deployed.  

I think 5 minutes is a good number because it's slightly longer than the average time on site but not so long that it will mess with developers who might wonder why changes they deployed to studio sites aren't working on their friends phones etc.


b1tr0t:
@jansepar Can you have a look at this?

jansepar:
@b1tr0t sorry I haven't had a chance to take a look yet. I will have a quick look today and give it a thorough review tomorrow!

b1tr0t:
@johnboxall We need at least two pairs of eyes on a critical path commit like this -- could you have a look as well?

jansepar:
@b1tr0t one thing that makes this pull request hard to review is that the original file was never in the repo in the first place, and thus I can't really see what changed... Do you think you could take that original file and commit it directly to master so we can see the diff?

jansepar:
I feel like we can safely close this - its been open for 1 year and 10 months :)

PR #94, opened 2013-02-13
V2.0

jansepar:
!!! DON'T MERGE SPACE COWBOYS !!!

Opening up this pull request to start discussion of making v2.0 the default library. There are a number of outstanding issues before this goes live:
- Create workflow with mobify-client for creating projects that use templating, or other custom libraries
- Documentation
- More examples


Makefile
```diff
@@ -1,9 +1,11 @@
 install:
-	git submodule update --init
+	#git submodule update --init
+	npm uninstall -g grunt
```
noahadams, line 3: Running a global uninstall in the makefile is probably bad for general use. Also, the -g installs are going to fail for people not running as root.

examples/framework/main.js
```diff
@@ -0,0 +1,73 @@
+require.config({
+    "paths": {   
+        "Mobify": "http://cloud.dev:8000/static/mobifyjs/mobify"
```
noahadams: o rly?

examples/framework/main.js
```diff
@@ -0,0 +1,73 @@
...
+
+require(["Mobify", "mustache",], function(Mobify, mustache) {
+
+    console.log("Executing main during capturing phase!")
+    var $html = Mobify.Capture.getSourceDOM();
+
+    //$html.find("script").remove();
+    $html.find("img").resizeImages();
+
+    //Mobify.Lazyload.rewriteSrc($html[0]);
```
noahadams: These examples could use a little light cleanup and tightening up if they're going to stick around.

mobify-custom.js.example
```diff
@@ -0,0 +1,30 @@
+require(["capture", "resizeImages"], function(Capture, ResizeImages) {
+    var Mobify = window.Mobify = window.Mobify || {};
+
+    // === HACKS TO ENSURE NEW CODE WORKING WITH v6 TAG === //
```
noahadams: Should we worry about v6 tag compatibility in an example?

noahadams:
kill the old/ directory :smiling_imp: :japanese_ogre:

src/capture.js
```diff
@@ -0,0 +1,416 @@
+define(["utils"], function(Utils) {
+
+// ##
+// # Static Variables/Functions
+// ##
+
+var openingScriptRe = /(<script[\s\S]*?>)/gi;
+
+// Inline styles are scripts are disabled using a unknown type.
```
noahadams: style _and_ scripts

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+var tagDisablers = {
+    style: ' media="mobify-media"',
+    script: ' type="text/mobify-script"'
+};
+
+var tagEnablingRe = new RegExp(Utils.values(tagDisablers).join('|'), 'g');
+
+// Map of all attributes we should disable (to prevent resources from downloading)
+var disablingMap = { 
+    img:    ['src', 'height', 'width'],
```
noahadams: do we still want to disable width and height? We do it currently to get rid of intrinsic layout on the element when mobifying (and never actually re-enable them). Worth mentioning that iframes have width and height too....

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    iframe: ['src'],
+    script: ['src', 'type'],
+    link:   ['href'],
+    style:  ['media'],
+};
+
+var affectedTagRe = new RegExp('<(' + Utils.keys(disablingMap).join('|') + ')([\\s\\S]*?)>', 'gi');
```
noahadams, line 26: generally speaking, this style of regex building is a little risky as there could be regex significant characters in the strings we're shoving into the regex, though we're pretty explicit about that here

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    var div = document.createElement('div');
+    div.appendChild(el.cloneNode(true));
+    var contents = div.innerHTML;
+    div = null;
+    return contents;
+}
+
+/**
+ * Returns a string of the unescaped content from a plaintext escaped `container`.
+ */
```
noahadams, line 74: Make this comment more clear about how this is used, eg. to get the contents of head and misc other unescaped tags

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+            return '';
+        }
+        return el.outerHTML || el.nodeValue || outerHTML(el);
+    }).join('');
+}
+
+/**
+ * Transform a string <tag attr="value" ...></tag> into corresponding DOM element
+ */
+function cloneAttributes(sourceString, dest) {
```
noahadams: this is badly named, because you're parsing a string and not cloning an object.

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    var match = sourceString.match(/^<(\w+)([\s\S]*)$/i);
+    var div = document.createElement('div');
+    div.innerHTML = '<div' + match[2];
+
+    [].forEach.call(div.firstChild.attributes, function(attr) {
+        dest.setAttribute(attr.nodeName, attr.nodeValue);
+    }); 
+
```
noahadams: I'll grudgingly accept that this will "parse like the browser parses" but it seems expensive to create all these throwaway elements whilst parsing, Note from Shawn: re-use one div for this in a closed over scope.

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    return dest;
+};
+
+// ##
+// # Constructor
+// ##
+var Capture = function(_doc, prefix) {
```
noahadams: some unnecessary underscores happening

examples/framework/main.js
jansepar: This example is REAAAALLLY old. I should actually remove it.

examples/framework/main.js
jansepar: Yeah I definitely plan to clean up the examples

mobify-custom.js.example
jansepar: I set this up mainly for Peter to be able to generate custom builds that work with one of our clients existing tags.

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+
+                // Disable before and the <script> itself.
+                // parsed = [before, <script>, script contents]
+                parsed = fragment.split(openingScriptRe);
+                parsed[0] = parsed[0].replace(affectedTagRe, disableAttributes);
+                if (parsed[1]) parsed[1] = parsed[1].replace(affectedTagRe, disableAttributes);
+                return parsed;
+            });
+
+    return [].concat.apply([], ret).join('');
```
noahadams, line 162: What's with the application of concat here?

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    // During capturing, we will usually end up hiding our </head>/<body ... > boundary
+    // within <plaintext> capturing element. To construct source DOM, we need to rejoin
+    // head and body content, iterate through it to find head/body boundary and expose
+    // opening <body ... > tag as a string.
+
+    // Consume comments without grouping to avoid catching
+    // <body> inside a comment, common with IE conditional comments.
+    var bodySnatcher = /<!--(?:[\s\S]*?)-->|(<\/head\s*>|<body[\s\S]*$)/gi;
+
+    captured = Utils.extend({}, captured);
```
noahadams: wat? this looks like a no-op to me.

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    } else {
+        document.addEventListener('DOMContentLoaded', this.restorer, false);
+    }
+};
+
+/** 
+ * Gather escaped content from the DOM, unescaped it, and then use 
+ * `document.write` to revert to the original page.
+ */
+Capture.prototype.restorer = function() {
```
noahadams: restore would be a better name here, methods should be verby, not nouny.

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+    // Reconstruct html, body, and head with the same attributes as the original document
+    cloneAttributes(this.htmlOpenTag, htmlEl);
+    cloneAttributes(this.headOpenTag, headEl);
+    cloneAttributes(this.bodyOpenTag, bodyEl);
+
+    // Set innerHTML of new source DOM body
+    bodyEl.innerHTML = disable(this.bodyContent, this.prefix);
+    var disabledHeadContent = disable(this.headContent, this.prefix);
+
+    // On some browsers, you cannot modify <head> using innerHTML.
```
noahadams: Which browsers?

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+/**
+ * Rewrite the document with a new html string
+ */
+Capture.prototype.render = function(htmlString) {
+    var doc = this.doc;
+
+    // Set capturing state to false so that the user main code knows how to execute
+    capturing = false;
+
+    window.setTimeout(function(){
```
noahadams: Just wondering why we make this async?

src/capture.js
```diff
@@ -0,0 +1,416 @@
...
+
+/**
+ * Grab the source document and render it
+ */
+Capture.prototype.renderCapturedDoc = function(options) {
+    var doc = this.capturedDoc;
+
+    // Most (all?) non-webkit browsers remove objects after document.write,
+    // therefor we must re-inject the library into the newly rendered DOM.
+    // AKA, new Ark :)
```
noahadams: This comment is too mobify jargony, maybe have it read soemthing like "After document.open(), the new document will have a fresh, clean javascript namespace, so to provide our library functionality afterwards, we must re-inject the script, except on webkit, where the global namespace survives unharmed."

src/resizeImages.js
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+    }
+    return imgs;
+}
+
+var defaults = resizeImages.defaults = {
+        selector: 'img'
+      , attribute: 'x-src'
+      , projectName: ''
```
noahadams, line 88: when we release this we should add a module that reaches in and populates this in the presence of a project.json or similar.

noahadams:
OK, I think I'm done with my review for now. overall things seem pretty sound, for the most part things are well named and commented to the point of near-understandability. I heartily approve of what you've done here, those there's still a fair bit of spit and polish to add.

src/capture.js
jansepar: plaintext won't grab the content of the site until the next tick. I'll add a comment :)

src/capture.js
jansepar: I ran into the issue on Firefox 4.0.

src/capture.js
jansepar: There is already a restore function :). But they actually could probably just be packaged into one function.

src/capture.js
jansepar: lol. I think there was something else at one point for why this made sense. Good catch!

src/capture.js
jansepar, line 162: Not sure :)

jansepar, line 162: Not sure, this is untouched from previous code!

src/capture.js
jansepar: In my new branch i used a cached div element instead of recreating it each time. Tested it and it works fine.

src/capture.js
jansepar: Fair :)

src/capture.js
jansepar: Sorry, my previous answer was in regards to plaintext. The reason it's setTimeout here is because this method happens async. For historical purposes, I will leave it. We may want to look into performance impact of doing this though, to see if it's worth keeping.

noahadams:
But what if there is no master?

jansepar:
This was just open for code review! It can be closed.

PR #98, opened 2013-02-15
V2.0 smallfixes

jansepar:
Lots of fixes based on feedback from @noahadams and @b1tr0t !


jansepar:
Whoops, pull request should be against v2.0.

Issue #108, opened 2013-03-05
Increase the detail of error messaging in console


tedtate:
Currently when you encounter an error when making a selection inside of your mobify.konf you get an error message that looks something like this:

![Developer_Tools_-_https___mygreen garmin com_en-US_US_on-the-road_automotive_n-uuml-vi-reg-40_prod92773 html-2](https://f.cloud.github.com/assets/136145/219124/069d7bc4-8521-11e2-8eba-5fee28c48270.png)

It would be nicer if the error was:

a) A bit louder (logged to console.error instead of console.warn)
b) Showed more detail about the error instead of just 'ReferenceError {}' or 'SyntaxError {}'

![Developer_Tools_-_https___mygreen garmin com_en-US_US_index ep mobify-override_mobify-path true_mobify-url http___localhost_8080__mobify-domain _mobify-all true_mobify 1_mobify-debug 1_mobify-js 1](https://f.cloud.github.com/assets/136145/219668/24fa0d9c-853c-11e2-9e22-e16c4585a759.png)

PR #109, opened 2013-03-05
Call toString on Error objects when logging to console.

tedtate:
Proposed fix for #108.

Tested in Chrome 27 and Safari 6.

Your thoughts @noahadams @rrjamie?


johnboxall:
:+1:

PR #112, opened 2013-03-05
fixes combo warm cache test by re-ordering cache loading and combineScripts() call

noahadams:
..., it is unclear to me why this has been passing with phantomjs

### DIFF
--- a/tests/fixtures/push.js
+++ b/tests/fixtures/push.js
@@ -1,0 +1,5 @@
+if(a.length == 1) {
+    a.push("b");
+} else {
+    a.push("d");
+}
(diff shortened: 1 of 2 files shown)


b1tr0t:
@noahadams Was there actually a regression of some kind on warm cache?  Or did someone just notice the test was passing and shouldn't have?

noahadams:
I noticed that this was consistently failing in chrome, but with good reason (the main thing is the call to combineScripts before the cache had been reset), and that there was a failing request happening as a result (as the fallback for combo execution when the script is not present in its resources dictionary is to write out a script tag referencing the original src).

PR #113, opened 2013-03-08
Run make install before tests

MikeKlemarewski:
Run make install before tests run.  This initializes the zepto submodule and makes sure tests don't barf.


noahadams:
+1 runs correctly or circle, thanks Mike!

PR #117, opened 2013-03-20
V2.0 swift jazzcat merge

jansepar:
!!! DO NOT MERGE SPACE CADET !!!

This pull request is open to review the differences between new jazzcat and old jazzcat.


jansepar:
Thanks @romanrudenko for your efforts on the updated Jazzcat! I've also brought in all of the tests and hacked them together to work with 2.0, so we have just as much coverage as we had before. This is ready to merge into the swift branch. :+1:

src/capture.js
```diff
@@ -354,6 +354,12 @@ Capture.prototype.escapedHTMLString = function() {
  * Rewrite the document with a new html string
  */
 Capture.prototype.render = function(htmlString) {
+    if (!htmlString) {
```
johnboxall, line 357: @jansepar hoist this `var` declaration?

jansepar, line 357: TOO BAD ITS ALREADY MERGED SUCKAAA. lol jk sure I'll make the change in the branch that this has been merged into.

PR #119, opened 2013-03-28
V2.0 swift jazzcat project parameter

noahadams:
Adds project slug parameter to generated urls where projectName is set of Jazzcat.combineScripts.defaults, and related tests


jansepar:
:+1: Thanks Noah! Also fixed up a few things in our examples, and in our package.json

PR #123, opened 2013-04-04
change hosts for jazzcat and image resizer to jazzcat.mobify.net and ir....

noahadams:
...mobify.net, respectively, new happy fun cookie free domains


jansepar:
Is this change safe to merge in yet?

noahadams:
The services are up and running on those domains, should be good, though I just realized I might need to update some tests.

jansepar:
The changes aren't in the actual build file, could you run `grunt build` to update those files? 

Also I'm still not 100% certain on this flow of committing the build files (if we get rid of them we'll have to at least create tags for each version release). But for now let's just roll with it.

noahadams:
I'll do a build and commit, but you're right, there's no "release process", but then again, these aren't hitting master either. At some point we can maybe steal the tiny bit of release automation from the modules, but we'll want ot have some policy on versioning and stuff aside from working in feature branches forever and submoduling them around.

jansepar:
Tried it, works. :+1: Thanks @noahadams!

PR #126, opened 2013-04-08
V2.0 remove mobify net hostnames

noahadams:
!! CAREFUL THERE, SPACE COWBOY !!

Reverts jazzcat and image resizing back to the hosts jazzcat.mobify.com and ir0.mobify.com because the cookie-free versions on mobify.net do not have SSL up on them yet.

This would re-introduce the mobify.com cookie problem in exchange for getting working SSL.


jansepar:
:+1: works

PR #128, opened 2013-04-09
adds use of project name parameter to mobify 1.1 jazzcat implementation,...

noahadams:
... also refactors the url generating function and adds a test for it

### DIFF
--- a/api/combo.js
+++ b/api/combo.js
@@ -275,9 +275,12 @@ var $ = Mobify.$
   , defaults = combineScripts.defaults = {
         selector: 'script'
       , attribute: 'x-src'
-      , endpoint: '//jazzcat.mobify.com/jsonp/'
+      , proto: '//'
+      , host: 'jazzcat.mobify.com'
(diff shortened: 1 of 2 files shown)


api/combo.js
```diff
@@ -365,8 +371,11 @@ $.fn.combineScripts = function(opts) {
     return combineScripts.call(window, this, opts)
 }
 
+// expose defaults for testing
+$.fn.combineScripts.defaults = combineScripts.defaults;
+
 Mobify.cssURL = function(obj) {
-    return '//combo.mobify.com/css/' + JSONURIencode(obj)
+    return '//jazzcat.mobify.com/css/' + JSONURIencode(obj)
```
rrjamie, line 378: If we are making `combineScripts.defaults` configurable, shoudn't we make `css` configurable?

noahadams, line 378: We should probably make css stuff a real API first.

jansepar:
@tedtate could you have a look at this?

PR #129, opened 2013-04-11
V2.0 readystate bug

jansepar:
- Bumps version from alpha1 to alpha2.
- Ensures that we don't capture until the document is in either complete, interactive, or loaded state.
- Updates documentation and examples accordingly.

### DIFF
--- a/src/capture.js
+++ b/src/capture.js
@@ -94,7 +94,7 @@ var cachedDiv = document.createElement('div');
 // # Constructor
 // ##
 var Capture = function(doc, prefix) {
-    this.doc = doc || document;
+    this.doc = doc;
     this.prefix = prefix || "x-";
 
     var capturedStringFragments = this.createDocumentFragmentsStrings();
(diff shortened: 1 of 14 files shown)


noahadams:
This mostly looks good, fix up the docs and I'll give you my approval

src/capture.js
```diff
@@ -104,8 +104,28 @@ var Capture = function(doc, prefix) {
...
+    }
+    // iOS 4.3, some Android 2.X.X have a non-typical "loaded" readyState,
+    // which is an acceptable readyState to start capturing on, because
+    // the data is fully loaded from the server at that state.
+    if (/complete|interactive|loaded/.test(doc.readyState)) {
+        createCapture(callback, doc, prefix);
+    }
+    else {
+        var created = false;
+        doc.addEventListener("readystatechange", function() {
```
noahadams, line 126: maybe a check of the readystate here? or at least a comment asserting what the readystate should be? This was confusing enough that it should be documented.

noahadams:
:+1:

src/capture.js
```diff
@@ -104,8 +104,32 @@ var Capture = function(doc, prefix) {
...
+    if (/complete|interactive|loaded/.test(doc.readyState)) {
+        createCapture(callback, doc, prefix);
+    }
+    // We may be in "loading" state by the time we get here, meaning we are
+    // not ready to capture. Next step after "loading" is "interactive",
+    // which is a valid state to start capturing on, and thus when ready
+    // state changes once, we know we are good to start capturing.
+    else {
+        var created = false;
+        doc.addEventListener("readystatechange", function() {
```
johnboxall, line 126: @jansepar Why'd you decide to go with `readystatechange` here and not `DOMContentLoaded`?

jansepar, line 126: Either could have been used. DOMContentLoaded is the event only gets fired when going from "loading" to "interactive", where as "readystatechange" is fired upon any change of the ready state.

shogun70, line 126: In my testing all IE versions will have `document.readyState == "interactive"` almost from the top of the document. If a script at the top of the page does manage to detect the `loading` state then almost immediately there will be a `readystatechange` event. I could only detect this initial change when refreshing a page, but the `interactive` state is consistent. 

I believe `document.readyState == "complete"` is reliable and should be early enough for `<plaintext>` capturing because there are no resources to download and no page to be painted.

PR #132, opened 2013-04-16
V2.0 image resize docs

jansepar:
Finally docs for Image Resizer! @noahadams could you review plz?


www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+* TOC
+{:toc}
+
+## `ResizeImages.resize(imgElements, [options])`
+
+__imgElements__ must be an array of image elements to resize.
+
+__options__ are optional.
+
+Rewrites the `src` of every image on the page based on the options
```
noahadams: every image in the array `imgElements`

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+passed. By default, images are requested through `ir0.mobify.com`,
```
noahadams: ...through mobfy's image resizing web service, `ir0.mobify.com`,...

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+maximum width is determined by the width of the device, format of
```
noahadams: maximum dimensions are based on the size of the device, taking into account is pixel density.

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+the image is maintained, and the image is cached forever.
```
noahadams: the requested image is indefinitely cacheable.

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+
+**Options**
+
+- `attribute`: `img` resource attribute to modify. Defaults to "x-src". "x-" is the default escape prefix used in [Capturing](/mobifyjs/v2/docs/capturing/)
```
noahadams: s/resource/element

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+- `projectName`: The project slug of the project on Mobify Cloud. Defaults to ""
+- `cacheHours`: Sets the length of time for the image(s) to be cached on the CDN. Defaults to forever.
+- `format`: Format of the image(s) being resized. Defaults to original format, except gifs, which are converted to pngs.
```
noahadams: `format`: Output format...

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+- `maxWidth`: Width of the image(s) being resized. Defaults to automatically determine width of device.
```
noahadams: _maximum_ width (does not apply to images that are already smaller than that width), also state that this parameter is given in css pixels, and device pixels will be computed based ont he device-pixel-ratio

www/v2/docs/image-resizer.md
```diff
@@ -0,0 +1,77 @@
...
+        var capturedDoc = capture.capturedDoc;
+        // Resize images using Mobify Image Resizer
+        var images = capturedDoc.querySelectorAll('img');
+        Mobify.ResizeImages.resize( images, {
+            projectName: "mobifytest",
+            cacheHours: "2",
+        } );
+        capture.renderCapturedDoc();
+    });
+
```
noahadams, line 55: Another good example might be resizing images to a specific size if a certain breakpoint is met, e.g. resize to 500px wide if device-width is between 768 and 1024.

noahadams:
I think I'm done making a ton of line comments, please ask if you'd like any clarification.

www/v2/docs/image-resizer.md
jansepar: Ah, this param was always ment to take the css pixels as an argument? I actually thought that was a bug and fixed that (it's in this PR). Could you double check my work? When you're done looking, poke me and let's have a chat about the right way to go about doing it.

noahadams: Yeah, the original API was designed to be used explicitly, by designers/developers, when writing transformations for several devices that shared common layouts (jn css pixels), but might have more or fewer device pixels (eg. iPhone 3Gs and 4, iPad 2 and 3).

src/resizeImages.js
```diff
@@ -106,10 +106,12 @@ ResizeImages.resize = function(imgs, options) {
...
     var width = opts.maxWidth || screenSize.width;
     var height = opts.maxHeight || screenSize.height;
-    if (dpr) {
-        opts.maxWidth = Math.ceil(width * dpr);
-        opts.maxHeight = Math.ceil(height * dpr);
+    if (dpr && !opts.maxWidth && !opts.maxHeight) { 
+        width = width * dpr;
+        height = height * dpr;
     }
+    opts.maxWidth = Math.ceil(width);
```
noahadams, line 119: this should be in css pixels :)

tests/resizeImages.html
```diff
@@ -45,12 +70,55 @@
...
+            });   
             var src = $('#resizeImages-non-http img').attr('x-src');
 
             equal(src, "gopher://archie.ftp.mailto/somanyprotocols", "non-http img should not resize");
         });
 
+        test('resize images - high dpr', function() {
+            var images = document.querySelectorAll('#resizeImages-high-dpr img');
+
+            window.devicePixelRatio = 2;
```
johnboxall: @jansepar Typically when setting a global variable like this in a test it's common convention to use a `try { ... } finally { ... }` block to make sure it is reset to it's original value. You can see a similar example in Zepto.js: https://github.com/madrobby/zepto/blob/master/test/zepto.html#L1340

www/v2/docs/image-resizer.md
jansepar, line 55: Added!

src/resizeImages.js
```diff
@@ -95,22 +93,32 @@ var getImageURL = ResizeImages.getImageURL = function(url, options) {
  * resized.
  */
 ResizeImages.resize = function(imgs, options) {
-    var opts;
+    var opts = Utils.clone(defaults);
```
noahadams, line 96: we might want to improve the extend implementation in utils to include a target and two sources ala jquery to obviate the need for the clone here, but that's not a blocker.

jansepar, line 96: Agreed!

noahadams:
this looks good, :+1:

Issue #137, opened 2013-04-28
Breaks browser history on IE, Firefox, Opera


shogun70:
On IE, Firefox and Opera, the `document.open(); document.write(); document.close()` call seems to add a new browser history entry with the same URL. 

After this occurs, clicking the browser back-button will return to the previous history item which is still the current URL. 

On IE and FIrefox this results in the current page reloading, the boot-script and mobify re-executing, `document.open()/write()/close()` being called and thus back to where we started. You can't leave the current page by clicking the back-button. 

Opera however, restores the state of the page before `open()/write()/close()` - this is just a blank page. Clicking the back-button once more will return to the real previous page.

jansepar:
Hi @shogun70, thanks for filing this! After doing some research, it seems like doing this fixes the back button for Firefox:

document.open("text/html", "replace");

I still have to do some testing on the other browsers you mentioned.

shogun70:
Thanks @jansepar. I did not know that. 

This fix seems fine for IE10, and I guess Opera can be ignored until the Webkit rebase - Opera 13??

shogun70:
@jansepar I saw that the patch for this issue only updates the `document.open()` call in `render()`. Shouldn't this also be done for `restore()`?

jansepar:
Yes. Yes it should :)

shogun70:
Also, there currently isn't a fix for Opera. Should the boot-script be changed to reject Opera?

jansepar:
I have a solution that would work specifically for Opera:

```
history.back = function(){history.go(-2)};
```

jansepar:
Hmmm, executing `history.back()` works, and I assumed that would override the back button, but that isn't the case.

jansepar:
Opera Next does not have the back button problem, now that it is powered by Blink. Since the back button is no longer a problem in IE, Firefox, and the latest Opera, I am closing this ticket.

PR #138, opened 2013-04-29
V2.0 back button

jansepar:
Fixes back button issue for IE10 and Firefox raised by issue https://github.com/mobify/mobifyjs/issues/137.

Started writing a regression test, but was thwarted by the fact that the `history` object is shared between parent and iframes, thus I could not simulate `history.back()`...

I also noticed when debugging this issue that we broke Mobify post-flood (post document.write). Not good! This was determined to be broken during the merge of the swift feature branch. This has been fixed, and a regression test has been added.

@noahadams could you review plz?


tests/mobify-library.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html class="foo">
+<head>
+  <meta charset="utf-8">
+  <title>capture.js tests</title>
```
noahadams: o rly?

noahadams:
The use of the symlink for the mobifyjs directory causes the tests not to be runnable natively on windows (though the capture.js, and mobify-library tests pass on IE10 when loaded _from_ an external machine that resolves the symlink properly.

Consequently, running `grunt test` on windows doesn't work either. 

I wouldn't characterize these as blockers however.

jansepar:
@noahadams added a note about the Windows issue it in the README, which should be sufficient for now.

tests/mobify-library.html
```diff
@@ -0,0 +1,50 @@
...
+<script>
+    // http://api.qunitjs.com/QUnit.config/
+    QUnit.config.autostart = false;
+    QUnit.config.testTimeout = 30 * 1000;
+
+    QUnit.start();
+
+
+    /* 
+     * All rendering engines (aside from Webkit) remove all objects
```
noahadams: Maybe more like "Most browsers create a new global javascript namespace in the new document, except webkit, which preserves the namespace as it was before the call to document.open()"

jansepar: Sure.

noahadams:
address my comment and your comment and you'll get your cherished +1

noahadams:
:+1:

PR #141, opened 2013-05-07
Update Unblockify

jansepar:
Updating unblockify to take a NodeList as an argument, instead of grabbing all scripts


noahadams:
made a few changes for style, but otherwise looks good, :+1:

PR #142, opened 2013-05-07
Localstorage cookie issue

jansepar:
Fixes issue with Jazzcat barfing due to localStorage being broken when cookies are disabled.


noahadams:
I'll counter-ask for a +1 here @jansepar

jansepar:
Looking good! Tested and works - :+1:

PR #143, opened 2013-05-09
IE10 tag fixes

jansepar:
- Fixes bootstrap tag to work in IE10. Done by binding the onload event before the script is attached to the document.
- Changes all examples to point to latest mobify.js build

Issue #127, opened 2013-04-09
Inserting <plaintext> means mobify.js AND main.js are single-points-of-failure


shogun70:
Once the `<plaintext>` tag has been inserted the page is blank until it is rewritten. 
This turns into a completely unusable page / site if anything prevents this rewrite from happening, e.g. 
- `mobify.js` fails to load because it is on an external server (CDN perhaps) which is down or blocked
- `mobify.js` throws an error, because it encounters an unexpected browser behavior
- `main.js` throws an error because it wasn't tested cross-browser

Ideally the initialization script provides a fallback to re-make the page as though there were no `<plaintext>` inserted.

johnboxall:
@shogun70 there are a couple of approaches to avoid this - assuming you have a modern browser, you could attach an error listener to the inserted script - on error you could set an "opt-out" cookie and reload the page.

shogun70:
An opt-out cookie throws the complexity back onto the site author. I think an option using the already loaded content would be more appropriate. 

Assuming the init script is the first script in the document, you could:
1. wait for `DOMContentLoaded` or equivalent
2. remove all script elements from document
3. `document.write` the `<html>` and `<head>` opening tags followed by `head.innerHTML` followed by the text-content of `<plaintext>`.

Something like:

```
onError = function() {
  var scripts = $$('script');
  for (var i=0, s; s=scripts[i]; i++) s.parentNode.removeChild(s);

  var head = $$('head')[0];
  var plaintext = $$('plaintext')[0];
  document.open(); 
  document.write(
    openingTag(document.documentElement), 
    openingTag(head),
    head.innerHTML,
    plaintext.innerHTML
  );
  document.close();
}

function openingTag(el) { 
  var tagName = el.tagName;
  return el.cloneNode(false).outerHTML.replace(RegExp('</' + tagName + '>$','i'), '\n');
}
function $$(tag) { 
  return document.getElementsByTagName(tag);
}
```

Probably a script loading timeout should also trigger this error recovery, so if `mobify.js` is taking a long time to download (because it is blocked) then it is abandoned and the unprocessed page is restored.

johnboxall:
@shogun70 Assuming the document is well formed and you aren't worried about losing content above bootstrap tag, your solutions should work!

https://gist.github.com/johnboxall/5355380

Some older browsers don't support `el.outerHTML` - depending on what browsers we're supporting, we may need to polyfill that.

[...]

PR #144, opened 2013-05-09
V2.0 tag optout

jansepar:
DO NOT MERGE ELSE I PITY THE FOOL!

TODO before merge: Once we're happy with the update to the tag, I'm going to update all the examples.

Forces opt out and reload when mobify.js fails to load.

Note: branched off of https://github.com/mobify/mobifyjs/pull/143

This PR addresses ticket #127


jansepar:
@johnboxall plz review!

tag/bootstrap.html
```diff
@@ -2,14 +2,15 @@
 (function(window, document, detector, mobifyjsUrl, mainUrl) {
 
 // If Mobify has been defined already, or if the detector fails,
-// exit
-if (this.Mobify || !detector()) {
+// or capturing is off, exit
+var optOut = /mobify-enabled=false/.test(location.hash + '; ' + document.cookie)
```
johnboxall: @jansepar Why'd you choose to use `mobify-enabled=false` and not something sorted like `mobify=0`?

tag/bootstrap.html
```diff
@@ -2,14 +2,15 @@
...
+if (this.Mobify || !detector() || optOut) {
```
johnboxall: @jansepar I'd move the check for `optOut` before `detector`.

tag/bootstrap.html
```diff
@@ -28,13 +28,25 @@
...
     var mobifyjsScript = document.createElement('script');
+    var errorHandler = function() {
+        var now = new Date();
+        // 30 minutes
+        var time = now.getTime() + 1800000;
+        now.setTime(time);
+        document.cookie = 'mobify-enabled=false' +
+                '; expires=' + now.toGMTString() +
+                '; path=/';
+        // Reload the page (location.reload fails in certain cases)
```
johnboxall: @jansepar what cases? Be explicit.

tag/bootstrap.html
```diff
@@ -28,13 +28,25 @@
     var Mobify = window.Mobify = window.Mobify || {};
     Mobify.capturing = true;
     var mobifyjsScript = document.createElement('script');
+    var errorHandler = function() {
+        var now = new Date();
```
johnboxall, line 36: @jansepar It's unclear to me how long to me how long session cookies last on most mobile browsers - so I think this could be a good compromise. I think you may be able to do this with a little less code:

```
(new Date((new Date) + 180000)).toGMTString()
```

tag/bootstrap.html
jansepar: I just wanted a more descriptive variable name, but I guess if you think of `mobify` as a verb, you can use it here by itself. Plus just using `mobify=0` keeps the tag small!

johnboxall: @jansepar if you use `mobify=0` and force int conversion, you can also use this as an "opt in".

jansepar:
@johnboxall could you have another look? Updated addressing your concerns.

examples/capturing-grumpycat/index.html
```diff
@@ -4,9 +4,13 @@
 <script>
 (function(window, document, detector, mobifyjsUrl, mainUrl) {
 
-// If Mobify has been defined already, or if the detector fails,
-// exit
-if (this.Mobify || !detector()) {
+// if mobify=(1|0) is set in hash or cookie, force mobify on/off.
+var match = /mobify=(\d)/.exec(location.hash + '; ' + document.cookie);
+if (match) {
+    if (!parseInt(match[1])) return;
```
johnboxall: @jansepar a few notes:
1. Because we're checking both the `location` and the `cookie` for this var, we might want to get a bit more specific: 

https://github.com/mobify/mobify-client/blob/master/vendor/tags/6/bootstrap.html#L35

This will help us avoid false matches if `mobify` somehow gets in the location or the cookie.
1. You probably don't need `parseInt` - just force conversion with `+`.
2. Why is the conditional on another line?

johnboxall:
@jansepar looking better. It's clear from this commit history that its very annoying to have to update the tag in multiple places. For the future, let's think about how we could move the tag definition to one source of truth.

examples/capturing-grumpycat/index.html
jansepar: Good call, I'll take that logic from the v6 tag. Already fixed the parseInt stuff :). And yeah, I can chop those two conditionals into one.

jansepar: Nvm, the conditional is on the second line because if there is an override, we no longer care about the detector.

jansepar:
@johnboxall updated again. Could you +1 if you're happy with it? I'll open a seperate PR updating all of the examples, and the quick start page.

jansepar:
And yeah, it sucks having to update all the examples when a tag change is made. I think the best solution here would be to template our examples and re-generate them when we make tag changes.

shogun70:
Is `location.hash` the best place for a user-requested opt-out? Alternatives:

a. the user can permanently set the `mobify` cookie
b. `sessionStorage` or `localStorage`
c. `location.search`

Options (a) & (b) would presumably be achieved with a `javascript:` link on the page. 

I guess option (a) is already enabled by this branch.

johnboxall:
@shogun70 - I think `location.hash` is the best place because:
1. It can easily be used in testing eg. append `#mobify=0` to the URL
2. It isn't sent to the server (some silly servers may choke on an extra query paramter)
3. It's widely supported

shogun70:
Sure. A JS bookmarklet to toggle the cookie is also easily used in testing.

johnboxall:
:+1:

PR #146, opened 2013-05-14
Better formatting and comments for the Jazzcat client.

johnboxall:
- Remove dead code
- Wrap a call that should be wrapped in a test


jansepar:
This is the one I meant to :+1:

PR #149, opened 2013-05-15
V2.0 nonbreaking jazzcat

jansepar:
@noahadams could you have a look?


src/jazzcat.js
```diff
@@ -230,10 +235,21 @@ define(["utils", "capture"], function(Utils, Capture) {
...
 
-    var absolutify = document.createElement('a');
-
-    var Jazzcat = window.Jazzcat = {
-        httpCache: httpCache
+    // Make Jazzcat for Firefox <= 11 and Opera 11/12 a noop.
+    // All other older Opera browsers are filtered out in the Mobify tag.
+    // These browsers have problems with document.write after a document.write
+    // Exposing on Jazzcat for use in qunit tests
+    Jazzcat.isIncompatibleBrowser = function() {
```
johnboxall: @jansepar if we're bothering to create this function, why don't we move the checks for `localStorage` and `window.JSON` here?

jansepar: Hmm, yeah thats probably a good point. I wanted a function that was mainly for checking the user agent, but I think moving `localStorage` and `window.JSON` into this method makes sense.

tests/jazzcat.html
```diff
@@ -108,6 +108,15 @@
                               .trim();
         };
 
+        var jazzcatNoop = Jazzcat.isIncompatibleBrowser();
+
+        test('isIncompatibleBrowser', function() {
+            ok(Jazzcat.isIncompatibleBrowser("Mozilla/5.0 (X11; U; Linux i686; ru; rv:1.9.1.3) Gecko/20091020 Ubuntu/10.04 (lucid) Firefox/4.0.1") === true, "Firefox 4.0");
```
noahadams: These could be fleshed out better, like, a few for browsers we think we'll never support (say, IE 6), a few for browsers we expect to have always supported (iOS 4.2) and then for the browsers where we support beyond a specific version, a failing test before the cusp and a passing one after.

Whaddaya think?

jansepar: So basically just more organization? Sounds good.

tests/jazzcat.html
```diff
@@ -253,6 +308,18 @@
         });
 
         asyncTest('httpCache - Least recently used eviction', 2, function() {
+            // Opt out of this test for Opera (which isn't compatible with
+            // jazzcat anyways) because the localstorage warning won't 
+            // dismiss on Saucelabs
+            if (jazzcatNoop) {
```
noahadams: maybe DRY these "make it pass" assertions up a bit?

jansepar: As in just throw the `if` around the tests?

noahadams: Or write a function that will run the requisite number of passing assertions, then call start(); in the async case

tests/jazzcat.html
```diff
@@ -108,46 +108,102 @@
...
+        });
+
+        if (Jazzcat.isIncompatibleBrowser()){
+
+            test('combineScripts - noop', function() {
+                var scripts = $($("#combineScripts-source").text()).filter("script").toArray();
+
+                var results = Jazzcat.combineScripts(scripts);
+
+                ok(!/Jazzcat/.test(results[0].innerHTML), "Jazzcat does not effect incompatible browsers");
```
noahadams: affect

noahadams:
Merged master in, this is... acceptable.

jansepar:
I'll take that as a :+1: :)

PR #153, opened 2013-05-16
Jazzcat loader regression

jansepar:
Fixes a regression for inserting the Jazzcat Loader, and adds a regression test.


build/mobify-2.0.0alpha3.js
```diff
@@ -1497,34 +1497,42 @@ define('jazzcat',["utils", "capture"], function(Utils, Capture) {
...
+      "\\('([\\s\\S]*?)'\\);<\\/script", "gi");
 
     /**
-     * Overrides `Capture.enable` to insert a Jazzcat bootloader to fetch all
-     * uncached scripts from the Jazzcat service before executing any Jazzcat calls.
+     * Insert the Jazzcat loader script before the first uncached script in
+     * some html markup string.
      */
-    Capture.enable = function() {
+    Jazzcat.insertJazzcatLoader = function(html) {
```
jansepar: Hmmm maybe this isn't the best name for this, since overriding `enable` is necessary for the insertion to happen. Gonna change it.

johnboxall:
@jansepar I've tweaked some things - can you review & merge?

jansepar:
@johnboxall so, the update that you made no longer tests the regression here. Jazzcat is broken because we changed the arguments in `Jazzcat.combo.exec`, but didn't change the regular expression to match it. Just testing that we are inserting some strings into the markup isn't enough to catch the regression.

jansepar:
:+1:

PR #155, opened 2013-05-17
Adds the ability to override the isStale behaviour

noahadams:
to be less http caching header compliant, by passing an option in through a call to combineScripts


johnboxall:
@noahadams my understanding is the behaviour you're looking to add here is the ability to enforce a minimum threshold to cache resources with un or underspecified cache headers.

You're modifying `isStale` to consider resources that were retrieved within a time period fresh. Eg. if this resource was retrieved in the last 10 minutes, consider it fresh, regardless of its headers.

Is there any reason why you wouldn't just add a second parameter to `isStale` and let the calling code decide whether to use some global options or not?

noahadams:
I agree that it would be nice to factor it up.

There's only one calling site for isStale right now, in httpCache.load(), so we could make it something passed in at that point, but it's not an option that's easily cascaded down the full call stack, that is to say setting the override is going to be something with a global side-effect until it is unset. I can do this, but I'll have to assert conditions on the result of calling httpCache.load() (which is probably better anyways as an integration test).

noahadams:
Also, realized there are flood issues with this approach :\ Cascade of passing options around it is!

PR #156, opened 2013-05-18
Using grunt-contrib-uglify instead of require's built-in uglify.

jansepar:
Using grunt-contrib-uglify instead of require's built-in uglify. This will remove the almond.js header for the minified mobify.js files.


noahadams:
OK, this works, and I like the additional orthogonality, the only thing I'm missing is the built file sizes. Any thoughts @jansepar ? Otherwise I'll go ahead and merge this.

noahadams:
:+1:

PR #157, opened 2013-05-21
V2.0 split bootloader

jansepar:
One case I haven't optimized for is when the scripts are already cached and the bootloader just loads from localstorage. Currently, this will insert this bootloader into the page twice, when it really only needs to be done once. 

I have an idea on how to handle this but I would love some input!

@noahadams please review!


src/jazzcat.js
```diff
@@ -270,11 +270,18 @@ define(["utils", "capture"], function(Utils, Capture) {
      * Note that this only the first part of the Jazzcat transformation. The
      * bootloader script is inserted by the overriden `Capture.enabled` function.
      */
-    Jazzcat.combineScripts = function(scripts, options) {
+    Jazzcat.combineScripts = function(scripts, doc, options) {
```
johnboxall, line 273: @jansepar I think you're okay to break compatibility here and make `doc` is a required parameter. We've haven't announced this as a public API.

src/jazzcat.js
```diff
@@ -283,14 +290,18 @@ define(["utils", "capture"], function(Utils, Capture) {
...
         httpCache.load();
-
         while (script = scripts[i++]) {
             url = script.getAttribute(options.attribute)
             if (!url) continue;
             script.removeAttribute(options.attribute);
             absolutify.href = url;
             url = absolutify.href;
-            script.innerHTML = !!httpCache.get(url) + ',' + options.execCallback + "('" + url + "');";
+            script.innerHTML = !!httpCache.get(url) +
```
johnboxall: @jansepar I think this line could benefit from either a comment, or be split into multiple assignments that reveal the purpose of what we're doing here.

src/jazzcat.js
jansepar, line 273: My thoughts were that if you wanted to use the document instead of a capturedDoc for whatever reason, you could just leave it blank. But it's actually probably better to enforce it to be a required parameter, incase people using the API in the future leave it blank, causing nothing to get concatenated.

tests/jazzcat.html
```diff
@@ -222,20 +223,28 @@
...
+                    doc.body.appendChild(scripts[2]);
+
+                    var combine = doc.getElementsByTagName("script");
+                    var results = Jazzcat.combineScripts(combine, doc);
+                    var html = Utils.outerHTML(doc.documentElement);
 
-                    var result = Jazzcat.insertLoaderIntoHTMLString(html);
-                    var expected = '<script src="//jazzcat.mobify.com/jsonp/Jazzcat.combo.load/%5B%22http%3A%2F%2Funcached%2F%22%5D"><\/script><script>true,Jazzcat.combo.exec(\'http://cached/\');<\/script><script>false,Jazzcat.combo.exec(\'http://uncached/\');<\/script>';
-                    equal(result, expected, "Loader inserted successfully")
+                    html = Jazzcat.insertLoaderIntoHTMLString(html, "head");
```
johnboxall: @jansepar What do you think about making two assertions here? Once after each run of `insertLoaderIntoHTMLString`? That way you can ensure that each call does what you think it does.

johnboxall: @jansepar There is a really a bit more to test here as well - if there are no scripts in the `<head>`, is a loader still inserted? (what about for the body)?

src/jazzcat.js
```diff
@@ -440,14 +459,18 @@ define(["utils", "capture"], function(Utils, Capture) {
...
     /**
      * Overrides `Capture.enable` to insert a Jazzcat bootloader to fetch all
      * uncached scripts from the Jazzcat service before executing any Jazzcat calls.
      */
     var oldEnable = Capture.enable;
     Capture.enable = function() {
-        var htmlStr = oldEnable.apply(Capture, arguments);
-        return Jazzcat.insertLoaderIntoHTMLString(htmlStr);
+        var html = oldEnable.apply(Capture, arguments);
+        // Insert seperate loaders for head and body
```
johnboxall: @jansepar last comment - you probably want to reference why this is a good idea.

src/jazzcat.js
jansepar, line 273: This is done.

jansepar:
@johnboxall added a bunch more tests for inserting the bootstrap loader! 4 scenarios in total.

src/jazzcat.js
```diff
@@ -409,45 +420,66 @@ define(["utils", "capture"], function(Utils, Capture) {
...
-      "\\('([\\s\\S]*?)'\\);<\\/script", "gi");
+    
+    /**
+     * Regex generator used to match Jazzcat calls in an HTML string.
+     * Generates regexp based on parent, which should either be head or body.
+     */
+    var execReGenerator = function(parent) {
+        return new RegExp("<script[^>]*?>(true|false),['\"]" +
+            parent + "['\"]," +
+            defaults.execCallback.replace(/\./g, '\\.') +
```
noahadams, line 435: This escaping should be more generic, it should catch and replace any regex significant character.

jansepar, line 435: Probably - but this isn't something I changed in this commit. I wouldn't say it's a blocker for this PR. If you do want to quickly add it though, seems like this regex should do us well: http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex

noahadams:
As discussed, my main blocking concern here is properly generating a nil bootloader (that is to say, doing nothing) for the case of body scripts that are already cached.

PR #159, opened 2013-05-24
WEBP support

jansepar:
DO NOT MERGE BULL ASTRONAUT!

This PR is to open a discussion about how we want to add webp detection to Mobify.js. The detection mechanism used by Modernizr seems to work well, but it is async. There are two options we can go with:
1. Default webp to false in the `defaults` object, and do the detection when loading the library, overriding `defaults` if there is support. Since we have to finish capturing before running the `resize` method, the feature detect should be complete by then.
2. Change the `resize` method to be async and do the detection there. The downside of this is having to require people to run `Capture.render` in the `resize` callback.
3. ?

Also please note webp is not live on the image resize service, so this PR will give you broken images :)


jansepar:
@noahadams @johnboxall could you have a look?

build/mobify-2.0.0alpha3.js
```diff
@@ -1009,12 +1009,87 @@ function getPhysicalScreenSize(devicePixelRatio) {
     }
 
     return multiplyByPixelRatio(sizes);
+}
+
+function persistWebpSupport(supported) {
```
noahadams, line 1035: I would strongly suggest preferring local storage to a cookie, cookies add extra bytes to every request against the domain, and can potentially be deleted server side (lots of european web servers will actively expire cookies they don't recognize these days).

build/mobify-2.0.0alpha3.js
```diff
@@ -1009,12 +1009,87 @@ function getPhysicalScreenSize(devicePixelRatio) {
...
+    function regexDetect(userAgent){
+        // Credit to Ilya Grigorik for WEBP regex matching
+        // https://github.com/igrigorik/webp-detect/blob/master/pagespeed.cc
+        var supportedRe = /(Android\s|Chrome\/|Opera9.8*Version\/..\.|Opera..\.)/i;
+        var unsupportedVersionsRe = new RegExp('(Android\\s(0|1|2|3)\\.)|(Chrome\\/[0-8]\\.)' +
+                                    '|(Chrome\\/9\\.0\\.)|(Chrome\\/1[4-6]\\.)|(Android\\sChrome\\/1.\\.)' +
+                                    '|(Android\\sChrome\\/20\\.)|(Chrome\\/(1.|20|21|22)\\.)' + 
+                                    '|(Opera.*(10|11)\\.)', 'i');
+
+        // Return false if browser is not supported
```
noahadams: actually return true and false here!

build/mobify-2.0.0alpha3.js
```diff
@@ -1009,12 +1009,87 @@ function getPhysicalScreenSize(devicePixelRatio) {
...
+    persistWebpSupport(support);
+
+    return support;
+
 };
 
 /**
  * Returns a URL suitable for use with the 'ir' service.
- */ 
-var getImageURL = ResizeImages.getImageURL = function(url, options) {
```
noahadams, line 1017: Why remove a module level name for this function?

noahadams:
gosh darn it, meant ot make those comments against something other than the built file...

noahadams:
also, really into the detailed testing

src/resizeImages.js
```diff
@@ -51,10 +51,85 @@ function getPhysicalScreenSize(devicePixelRatio) {
...
+
+    if (options) {
+        Utils.extend(opts, options);
+    }
+
+    if (!opts.disablePersist) {
+        // Check if WEBP support has already been detected
+        var match = /webpSupport=([^&;]*)/.exec(document.cookie);
+        if (match && match[1] == "0") {
+            return 0;
```
noahadams: on browsers where we have persisted false, we might want to opportunistically (and asynchronously) look for suport in the future, maybe we should record the test date and re-run the test if it's been a while?

build/mobify-2.0.0alpha3.js
jansepar, line 1035: I didn't use localstorage because of the lack of sharing between protocols and subdomains, but the points you listed are valid concerns.

build/mobify-2.0.0alpha3.js
jansepar, line 1017: So that it can be overridden by users like `ResizeImages.getImageURL = blah`.

src/resizeImages.js
jansepar: Any thoughts on how long a "while" should be?

noahadams: a month? a week? something in that range. Any opinions yourself? Imagine a site that's running on this exact code ina  year and a half, and has iphone visitors, and suddenly, miraculously, there is webp support in Safari iOS 7.2. What's the path for getting a regular visitor to a site to start getting webp images?

jansepar: I think a week is fine. It is a very small calculation anyways so we could probably get away with a day.

noahadams:
This looks like its nearly in shape!

noahadams:
you have my :+1:

PR #160, opened 2013-05-24
V2.0 fix bootloaders

noahadams:
Changes bootloader generation so that localStorage cache is loaded explicitly before the first jazzcatted resource, and requests to the service are made before the first uncached resource.


jansepar:
Added a test and fixed a comment. Otherwise looks good! :+1:

PR #161, opened 2013-05-24
V2.0 jazzcat refactor

jansepar:
Includes updates insert loaders into head and body, as well as fixes double loading.


jansepar:
:+1:

PR #162, opened 2013-05-28
V2.0 jazzcat cache override

noahadams:
Adds a "cacheOverrideTime" parameter to Jazzcat.combineScripts()'s options object that allows one to override the behaviour of isStale when loading scripts.


tests/jazzcat.html
```diff
@@ -215,6 +361,60 @@
...
+                    var doc = document.implementation.createHTMLDocument("");
+                    
+                    doc.write("<html><head><script src=\"/build/mobify.js\"><\/script></head><body>" + scriptText + "</body></html>");
+
+                    var scripts = doc.querySelectorAll('script');
+                    Jazzcat.combineScripts(scripts, doc, {cacheOverrideTime: 10});
+
+                    // verify that things we wanted to load from cache did and other things did not
+
+                    var enabledDocumentString = Capture.enable(Utils.outerHTML(doc.documentElement));
```
jansepar, line 405: Not a huge deal but curious why you used Capture.enable here, instead of just leaving `x-` inside of `expectedHTML`?

noahadams, line 405: I wanted to see that the correct bootloader was generated in the pass through enable. We could also unit test the functions this depends on.

jansepar, line 405: Ohhhh okay, I was thinking that you use enable to remove the `x-`. For most of my tests I just used `insertLoadersIntoHTMLString` to test the bootloader insertion, but I guess it's good to have at least one test using `enable` to ensure that method doesn't regress.

src/jazzcat.js
```diff
@@ -270,7 +287,12 @@ define(["utils", "capture"], function(Utils, Capture) {
      * Note that this only the first part of the Jazzcat transformation. The
      * bootloader script is inserted by the overriden `Capture.enabled` function.
      */
+
     Jazzcat.combineScripts = function(scripts, doc, options) {
+        if (options && options.cacheOverrideTime !== undefined) {
```
jansepar, line 304: Can we switch this to: `if (options && options.cacheOverrideTime) {`.

I believe it's more consistent with the style of our other `if` statements, and it's more legible.

src/jazzcat.js
```diff
@@ -396,7 +419,11 @@ define(["utils", "capture"], function(Utils, Capture) {
      */
     var _getLoadFromCacheScript = function() {
         var loadFromCacheScript = document.createElement('script');
-        loadFromCacheScript.innerHTML = "Jazzcat.httpCache.load();"
+        loadFromCacheScript.innerHTML = (httpCache.options.overrideTime ?
+          "Jazzcat.httpCache.load(" + JSON.stringify(httpCache.options) + ");" :
+          "Jazzcat.httpCache.load();" );
+
+        //loadFromCacheScript.innerHTML = "Jazzcat.httpCache.load();"
```
jansepar: Nuke this comment from orbit :)

src/jazzcat.js
noahadams, line 304: but what if you want to override it to 0? We could do some more serious validation as well, like, check the type and check that it's in an integer in [0, infty)

jansepar, line 304: Yeah I guess that should be possible (although not likely). Fair enough!

src/jazzcat.js
```diff
@@ -270,7 +287,12 @@ define(["utils", "capture"], function(Utils, Capture) {
      * Note that this only the first part of the Jazzcat transformation. The
      * bootloader script is inserted by the overriden `Capture.enabled` function.
      */
+
     Jazzcat.combineScripts = function(scripts, doc, options) {
```
jansepar, line 303: Could you throw into the comments above here what the options are that you can pass in?

src/jazzcat.js
```diff
@@ -58,9 +60,14 @@ define(["utils", "capture"], function(Utils, Capture) {
     /**
      * Load the cache into memory, skipping stale resources.
      */
-    var load = function() {
+    var load = function(options) {
         var data = localStorage.getItem(localStorageKey)
         var key;
+        var staleOptions;
+
+        if (options && options.overrideTime) {
```
jansepar: Should probably check for undefined here if we are being consistent with `combineScripts`.

src/jazzcat.js
```diff
@@ -176,13 +183,22 @@ define(["utils", "capture"], function(Utils, Capture) {
...
         var cacheControl = headers['cache-control'];
         var now = Date.now();
         var date;
+        var overrideTime;
+
+        // If a cache override parameter is present, see if the age of the 
+        // response is less than the override, cacheOverrideTime is in minutes, 
+        // turn it off by setting it to false
+        if (options && (overrideTime = options.overrideTime) &&
+          (date = Date.parse(headers.date))) {
```
jansepar: Why throw `date = Date.parse(headers.date)` into the `if`? What's the condition we are checking that might opt us out here?

src/jazzcat.js
```diff
@@ -176,13 +183,22 @@ define(["utils", "capture"], function(Utils, Capture) {
...
+            return (now > (date + (overrideTime * 60 * 1000)));
+        }
 
-        // If `max-age` and `date` are present, and no other no other cache
+        // If `max-age` and `date` are present, and no other cache
         // directives exist, then we are stale if we are older.
         if (cacheControl && (date = Date.parse(headers.date))) {
```
jansepar: You already create this `date` variable, don't need to parse again.

src/jazzcat.js
noahadams: the assignment in the if was a brevity thing for validating the parsing, but this can just get moved up.

src/jazzcat.js
noahadams: touché

src/jazzcat.js
```diff
@@ -269,8 +290,21 @@ define(["utils", "capture"], function(Utils, Capture) {
...
+     *
+     * - `cacheOverrideTime` :  An integer value greater than 10 that will 
+     *                          override the freshness implied by the HTTP 
+     *                          caching headers set on the reource.
      */
+
     Jazzcat.combineScripts = function(scripts, doc, options) {
+        if (options && options.cacheOverrideTime !== undefined) {
+            Utils.extend(httpCache.options,
+              {overrideTime: options.cacheOverrideTime});
```
jansepar, line 306: Don't we want to default the overrideTime to 10? Currently, this will cause a Jazzcat request (which will hit cache for 10 min), but I thought we decided to prevent that?

noahadams, line 306: The default of ten minutes is implemented directly in the isStale function, that's not an override.

src/jazzcat.js
```diff
@@ -173,18 +180,31 @@ define(["utils", "capture"], function(Utils, Capture) {
...
         var headers = resource.headers || {};
         var cacheControl = headers['cache-control'];
         var now = Date.now();
-        var date;
+        var date = Date.parse(headers['date']);
+        var overrideTime;
+
+        // Fresh if less than 10 minutes old
+        if (date && (now < date + 600 * 1000)) {
+            return false;
```
jansepar, line 195: Ah, I never saw this because I was looking at the next `if` statement. We could save ourself from running this logic twice if we just set `overrideTime = 10`. Or maybe something like this:

`cacheTime = options.overrideTime || 10`

noahadams, line 195: idk, I sort of like how the cascade of early-returning ifs that's already here reads: 

Fresh if newer than 10 minutes OR newer than override if present OR newer than http cache time if present
Otherwise, stale.

jansepar:
:+1:

PR #163, opened 2013-05-29
V2.0 image resize filter data uris

noahadams:
Adds a test to check that we don't try to resize data URIs, also some formatting cleanup (commas, semi-colons).


jansepar:
:+1: thanks for the test!

PR #164, opened 2013-05-30
V2.0 image api improvements

jansepar:
Various API improvements.
- ~~Get Resizer API to work without Capturing~~
- Integrate modified Picture Polyfill


build/mobify-2.0.0alpha3.js
```diff
@@ -1079,14 +1079,20 @@ ResizeImages.resize = function(imgs, options) {
     opts.maxWidth = Math.ceil(width);
     opts.maxHeight = Math.ceil(height);
 
-    var attr;
```
noahadams, line 1082: could we set the default as the library loads, instead of setting it at function call time?

jansepar, line 1082: Yep!

jansepar:
Closing in favor of picturefill PR I'm about to open.

PR #166, opened 2013-06-05
onerror handler for resizer

noahadams:
The goal here is to make pages survive an outage of the image resizing service somewhat intact, though less performant.

So, there are three ways to attach an error handler to an image element
1. using `el.addEventListener` in js
2. setting the `el.onerror` property
3. adding the `onerror="..."` HTML attribute

I had an original implementation using 1., but realized there was going to be am race attaching the handler in the generated document, and that 3., however ugly, was a very simple workaround for this. The only substantive problem I can think of with it is overriding an existing onerror handler, though we could easily check for that and not do it.

Anybody have any serious opinions about this?

Also, if you want to give the demo a try I recommend using charles or /etc/hosts to point ir0.mobify.com to nowhere inparticular.


jansepar:
Have you had a chance to profile this to see what kinds of performance implications this will bring us? I'd like to know before merging.

noahadams:
Just did that now, there's no discernable difference running it across 176 elements, takes between 3ms and 8ms with and without this code.

tests/resizeImages.html
```diff
@@ -123,6 +131,35 @@
...
+        });
+
+        test('resize images - onerror handler restores src', function () {
+            var img = new Image()
+            img.setAttribute("src", "foo");
+            img.setAttribute("data-orig-src", "bar");
+
+            ResizeImages.restoreOriginalSrc({target: img});
+
+            equal(img.getAttribute("src"), "bar");
```
jansepar, line 158: We could probably go a step further to test that onerror gets triggered by running resize on an image with a src that doesn't resolve, and inserting that image into the DOM to trigger the error hander.

jansepar:
:+1: looks good thanks @noahadams !

PR #168, opened 2013-06-12
Better webp detection

noahadams:
- Excludes Android 4 native browser with RegEx test
- Uses a more modern webp data URI


noahadams:
also, this needs a build before merge

johnboxall:
@noahadams could you leave a note about where that WEBP image came from?

jansepar:
:+1:!

PR #169, opened 2013-06-13
Fixes an issue in Android 2.3 where JSON.parse-ing `null` throws an exception

jansepar:
@noahadams have a look please!


noahadams:
this looks good :+1:

PR #172, opened 2013-06-24
V2.0 css optimizer client

noahadams:
This is some client code for the upcoming css features of jazzcat, review would be appreciated.


src/utils.js
```diff
@@ -43,6 +43,26 @@ Utils.clone = function(obj) {
     return target;
 };
 
+// Some url helpers
+/**
+ * Takes a url, relative ro absolute, and absolutizes it relative to the current
```
jansepar: to*

src/utils.js
```diff
@@ -43,6 +43,26 @@ Utils.clone = function(obj) {
     return target;
 };
 
+// Some url helpers
+/**
+ * Takes a url, relative ro absolute, and absolutizes it relative to the current 
+ * document's location, with the assistance of an a element.
+ */
+var _absolutifyAnchor = document.createElement("a");
+Utils.absolutify = function(url) {
```
jansepar, line 52: Useful util method! We use this in `resizeImages.js` as well (and maybe `capture.js`?), could you update those libraries to use this method as well?

jansepar, line 52: Oh, and probably `jazzcat.js` :)

src/utils.js
```diff
@@ -43,6 +43,26 @@ Utils.clone = function(obj) {
...
+    _absolutifyAnchor.href = url;
+    return _absolutifyAnchor.href;
+};
+
+/**
+ * Takes an absolute url, returns true if it is an http/s url, false otherwise 
+ * (e.g. mailto:, gopher://, etc.)
+ */
+var _httpUrlRE = /^https?/;
+Utils.httpUrl = function(url) {
```
jansepar, line 62: Same comment as the one above: please use this in other libraries to DRY things up.

src/cssOptimize.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * cssOptimize - Client code to a css optimization service
+ */
+
+define(["utils"], function(Utils) {
+
+var CssOptimize = window.cssOptimize = {};
+
+var absolutify = document.createElement("a");
```
jansepar: Pew pew pew

src/cssOptimize.js
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Takes an original, absolute url of a stylesheet, returns a url for that
+ * stylesheet going through the css service.
+ */
+
+CssOptimize.getCssURL = function(url, options) {
```
jansepar: Wouldn't the convention here be `getCssUrl` ? We could just make it consistent with jazzcat and call it `getURL`

src/cssOptimize.js
```diff
@@ -0,0 +1,89 @@
...
+        url = Utils.absolutify(attributeVal);
+        if (Utils.httpUrl(url)) {
+            element.setAttribute(options.targetAttribute,
+                                 CssOptimize.getCssURL(url));
+        }
+    }
+};
+
+/**
+ * Takes an array-like object of `<link>` elements
```
jansepar, line 49: We don't do any guarding against non-link elements. Should we throw in return if element.nodeName !== 'LINK' ? We also don't do that for `combineScripts` in jazzcat...

src/cssOptimize.js
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+};
+
+/**
+ * An 'error' event handler designed to be set using an "onerror" attribute that
+ * will set the target elements "href" attribute to the value of its 
+ * "data-orig-href" attribute, if one exists.
+ */
+var restoreOriginalHref = CssOptimize.restoreOriginalHref = function(event) {
```
jansepar, line 71: We do the same thing for images. What do you think of making this a utility function to DRY things up and reduce LOC.

src/cssOptimize.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+/**
+ * Rewrite the href of a stylesheet referencing `<link>` element to go through 
+ * our service.
+ */
+CssOptimize._rewriteHref = function(element, options) {
+    var attributeVal = element.getAttribute(options.targetAttribute);
+    var url;
+    if(attributeVal) {
```
jansepar: `if (attributeVal) {`

tests/cssOptimize.html
```diff
@@ -0,0 +1,66 @@
...
+            var l = document.createElement('link'),
+                origHref = 'bar';
+            l.setAttribute('data-orig-href', origHref);
+            l.setAttribute('href', 'foo');
+
+            cssOptimize.restoreOriginalHref({target: l});
+
+            strictEqual(l.getAttribute('href'), origHref);
+        });
+    });
```
jansepar, line 85: Could you add a test for `_rewriteHref` and `optimize`?

src/cssOptimize.js
noahadams, line 49: look down 8 lines

jansepar, line 49: Hah. Yeah, not sure how I missed that :)

tests/cssOptimize.html
noahadams, line 85: optimize is the first test, adding one for rewriteHref

jansepar, line 85: Ah, I missed that. Could you check that first tests description to "cssOptimize - optimize" to keep things consistent?

jansepar:
Could you throw up an example here that I can use to test this bad boy out? Also since most of our APIs have 1 or more examples, it would be great to have one for cssOptimize as well.

src/cssOptimize.js
noahadams, line 71: we'd need to parameterize the attributes then, and invoke with the correct ones.

jansepar, line 71: Yep. Change it to something like `restoreOriginalAttr` or something.

noahadams:
@jansepar I've included an example, it doesn't show much visually but you can inspect to see where the stylesheet is coming from (and for now, notice that the error handler works, and if you run it locally against jazzcat-next, that localhost blacklisting works.

jansepar:
Looking good! Giving this a conditional :+1: pending a merge and deploy of the CSS Optimization code in Jazzcat.

noahadams:
Thanks!

PR #173, opened 2013-06-24
clean up attribute getting/setting code

noahadams:
Renames "attribute" and "setAttr" properties to sourceattribute and targetAtribute, and sets them up in a more consistent way.


jansepar:
:+1:

Issue #136, opened 2013-04-28
document.readyState == 'interactive'  is invalid DOM ready test in IE10


shogun70:
This relates to a bug in the v2.0 branch.

`/complete|interactive/.test(document.readyState)` and `document.onreadystatechange` are not reliable as DOM ready tests. All versions of IE can have `interactive` state before the DOM is fully loaded. Also, `readystatechange` can fire before the DOM is fully loaded. 

I have written an exploratory test which is sufficient to confirm this bug.

https://gist.github.com/shogun70/5388420

The jQuery project also has a bug report - 
[document ready is fired too early on IE 9/10](http://bugs.jquery.com/ticket/12282) 
and a "fix" - 
[sometimes IE10 and lower have document.readyState == "interactive" way too early](https://github.com/jquery/jquery/pull/901)

domReady also has a report and fix - 
[Add Guard for invalid readyState in IE9 and IE10](https://github.com/ded/domready/pull/17)

shogun70:
Also, Opera doesn't fire the `readystatechange` event on `document`. 

CORRECTION: document.onreadystatechange fires **after** window.onload. 
I've updated the test:
https://gist.github.com/shogun70/5388420

jansepar:
@shogun70 thanks for the test case! I will look into this and come up with the appropriate fix asap!

shogun70:
Unlike `DOMContentLoaded` and `readystatechange`, `window.onload` is supported on all recent and almost all older browsers. When **mobify.js** is able to run on older browsers then a `window.onload` solution won't need to be updated. 

Because there is no content, images or styles to download and / or render, I would assume that `window.onload` is not significantly later than any other event that could be used.

PR #177, opened 2013-06-27
Fixes capturing on IE10 for Lumia 920

jansepar:
Addresses this bug: https://github.com/mobify/mobifyjs/issues/136

Initially, I could not replicate the bug on IE10 in a Windows 8 VM, but now the bug has surfaced on IE10 mobile for Lumia 920. We used to check for "interactive" as one test to determine if the DOM was ready, but in some variations of IE10, "interactive" is an incorrect state to poll.

### DIFF
--- a/src/capture.js
+++ b/src/capture.js
@@ -111,16 +111,27 @@ var init = Capture.init = function(callback, doc, prefix) {
         var capture = new Capture(doc, prefix);
         callback(capture);
     }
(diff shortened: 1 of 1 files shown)


jansepar:
@noahadams could you have a look? Also, don't merge, once you give me a +1 I will update the build file. I've also run this on saucelabs, and manually on Android 2.3, and things are looking good.

noahadams:
Looks good. Solid catch, and congratulations on finding an existing solution.

:+1:

PR #180, opened 2013-07-12
V2.0 change bootstrap

jansepar:
Makes changes to the bootstrap tag that allow you to use EITHER a main.js file, or throw main inside of an anonymous function. Makes examples, tests, and the quick-start code much simpler.


examples/resizeImages-img-element/main.js
```diff
@@ -1,19 +1,2 @@
...
-        // Resize images using Mobify Image Resizer
-        var images = capturedDoc.querySelectorAll('img');
-        Mobify.ResizeImages.resize(images);
-
-        // Render source DOM to document
-        capture.renderCapturedDoc();
-    });
-
-}
+// Main executable can now be found inline inside of the index.html file
```
noahadams: just delete this file...

examples/resizeImages-picture-element/main.js
```diff
@@ -1,20 +1,2 @@
...
-        
-        // Render source DOM to document
-        capture.renderCapturedDoc();
-    });
-
-} else {
-    console.log("Executing main in post-capturing phase!");
-    // Execute polyfill on document
-}
+// Main executable can now be found inline inside of the index.html file
```
noahadams: delete all these files, there is no sense in keeping around a .js file that contains only comments

jansepar: They are all linked in the article on Mozilla Hacks. I'd rather people land on this and then know where to go, instead of landing on a 404.

noahadams: ugh, see if you can update the article with a link to a specific commit, or a cdn version we can leave in place intact maybe?

noahadams:
ok, this looks good, :+1:

PR #181, opened 2013-07-16
Fixes bug in Jazzcat that causes all resources to not be saved

jansepar:
Fixes a bug that will cause resources to not get saved because we will (incorrectly) attempt to evict the same resource over and over when the set of resources is too big, which will eventually reach the maximum number of attempts and not save the rest of the resources.


tests/jazzcat.html
```diff
@@ -113,6 +113,9 @@
         // Too big for localStorage in Chrome, Webkit, and Firefox
         var LONG_STRING = Array(10000000).join('1');
 
+        // Large, but small enough to fit in cache
```
noahadams: s/cache/localStorage

tests/jazzcat.html
```diff
@@ -287,6 +290,30 @@
                 }, 0);
             });
 
+            // regression test to ensure Jazzcat does not evict everything
```
noahadams: it doesn't evict things, it fails to save because it /doesn't/ evict things

tests/jazzcat.html
```diff
@@ -287,6 +290,30 @@
...
+                        }
+                    }
+
+                    httpCache.save(function(err) {
+
+                        httpCache.reset();
+                        httpCache.load();
+
+                        ok(!!httpCache.get('val1'), '`val1` was cached');
+
```
noahadams, line 311: This test should be side effect free, so, clean things up here.

noahadams:
:+1: pending a good run on sauce

jansepar:
Done without errors on Sauce :) Thx @noahadams!

PR #182, opened 2013-07-16
V1.1 jazzcat updates

noahadams:
Cleans up code slightly, fixes a bug and adds default ten minute caching and some metadata to the cache

### DIFF
--- a/api/combo.js
+++ b/api/combo.js
@@ -55,10 +55,7 @@ var get = function(key, increment) {
           , resource
           , attempts = 10
           , key
-          , serialized
-            // End of time.
-          , lruTime = 9007199254740991
-          , lruKey;
+          , serialized;
 
         for (key in cache) {
             if (cache.hasOwnProperty(key)) {
@@ -67,6 +64,9 @@ var get = function(key, increment) {
         }
 
         (function persist() {
+            // End of time.
+            var lruTime = 9007199254740991
+            , lruKey;
             setTimeout(function() {
                 try {
                     serialized = JSON.stringify(resources);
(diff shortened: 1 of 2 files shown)


api/combo.js
```diff
@@ -203,8 +209,8 @@ var ccDirectives = /^\s*(public|private|no-cache|no-store)\s*$/
             }
 
             // If `expires` is present, we are stale if we are older.
-            if (expires = Date.parse(headers.expires)) {
-                return now > expires;
+            if (date = Date.parse(headers.expires)) {
+                return now > date;
```
jansepar, line 213: I found this a bit confusing to read. Could we do something like:

```
if (headers.expires){
    return now > Date.parse(headers.expires);
```

This way, we would only create a date object of expires exists, instead of `Date.parse`-ing a undefined object.

api/combo.js
```diff
@@ -343,7 +351,8 @@ var $ = Mobify.$
 
             for (i = 0; i < resources.length; i++) {
                 resource = resources[i];
-                if (resource.status == 'ready') {
+               if (resource.status == 'ready') {
+                    resource.loadedTime = now;
```
jansepar, line 355: I don't see `loadedTime` used anywhere else in this library. Could you explain/comment on what this is used for?

api/combo.js
noahadams, line 213: Admittedly, it reeks of terse C

The current implementation also checks that it parses to something truthy, we'd need to further nest ifs to ensure parsing happens correctly, though we could pull the assignment out of the if condition.

if headers.expires doesn't exist we get
`if (Date.parse(undefined))` which is `if (NaN)` and NaN is falsey, also, if date is funky (i.e. not parseable as a date) we get `if (Date.parse("abc"))` which becomes `if (false)`

Oh, also, `Date.parse()` returns time in milliseconds, not a date object.

api/combo.js
noahadams, line 355: it's just some meta data to throw in in case we want it in the future.

api/combo.js
jansepar, line 213: Ah. Okay I can buy that :)

jansepar:
Looking good, passing tests! :+1:

PR #183, opened 2013-07-19
V2.0 jazzcat improvements

jansepar:
Alright this is a big one. Changes:
- Have the option to use Jazzcat without concatenation
- Use Jazzcat with a JS response type as well as JSON
- `combineScripts` -> `optimizeScripts`
- No longer insert loaders with regular expressions.

Also, this was branched off of this: https://github.com/mobify/mobifyjs/pull/180. Merge it to have a less cluttered comparison!


Gruntfile.js
```diff
@@ -4,6 +4,7 @@ var path = require('path');
 
 var LONG_CACHE_CONTROL = "public,max-age=31536000, s-maxage=900"; // one year
 var SHORT_CACHE_CONTROL = "public,max-age=300"; // five minutes
+var NO_CLIENT_CACHE = "max-age=0, s-maxage=31536000, no-store";
```
noahadams: The "no-store" cache control directive will cause _every_ cache not to store, so the s-maxage directive here will be effectively ignored:
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.2

jansepar: Yup, I was 90% sure that was the case but I wanted to try it out anyways, and it was confirmed that this was the case. I'll update this to reflect that.

src/jazzcat.js
```diff
@@ -91,39 +93,51 @@ define(["utils", "capture"], function(Utils, Capture) {
...
+        var resources;
         var key;
 
-        for (key in cache) {
-            if (cache.hasOwnProperty(key)) {
-                resources[key] = cache[key];
-            }
+        // prevent multiple saves before onload
+        if (!canSave) {
+            return;
```
noahadams: we should invoke "callback" with an error before returning in this case

src/jazzcat.js
```diff
@@ -91,39 +93,51 @@ define(["utils", "capture"], function(Utils, Capture) {
...
         // evict an item and try again. Use `setTimeout` to ensure the UI stays
         // responsive even if a number of resources are evicted.
         (function persist() {
             var store = function() {
+                var resource;
                 var serialized;
                 // End of time.
                 var lruTime = 9007199254740991;
                 var lruKey;
+                resources = resources || (function(){
```
noahadams: this should probably just be a call to the clone helper from utils

src/jazzcat.js
```diff
@@ -148,10 +162,15 @@ define(["utils", "capture"], function(Utils, Capture) {
                     return persist();
                 }
 
+                canSave = true;
                 callback && callback();
             };
-
-            setTimeout(store, 0);
+            if (document.readyState === 'complete') {
```
noahadams: there's an argument to be made for allowing you to save at `interactive` (and also `loaded` for older iOS. Imagine, for example, a big image gallery page that takes 10+ seconds to reach complete, and someone navigates away first. I know we can't reliably catch this on IE, but that's a browser bug.

src/jazzcat.js
```diff
@@ -163,7 +182,7 @@ define(["utils", "capture"], function(Utils, Capture) {
     /**
      * Returns a parsed HTTP 1.1 Cache-Control directive from a string `directives`.
      */
-    var ccParse = function(directives) {
+    httpCache.ccParse = function(directives) {
```
noahadams: this should be on httpCache.utils as well

src/jazzcat.js
jansepar: The reason I delay it until complete is because if we saved at interactive, it would still delay the onload event if I'm not mistaken.

noahadams: It might, but generally speaking the biggest delay to the firing of load is image loading, which is driven through something other than the UI thread, and this can happen in parallel to that

jansepar: Hmm, I could change the tests to measure DOMContentLoaded instead of onload, which will be the exact same time since there are no image assets in the page.

src/jazzcat.js
```diff
@@ -293,127 +301,111 @@ define(["utils", "capture"], function(Utils, Capture) {
...
-        attribute: 'x-src',
-        base: '//jazzcat.mobify.com',
-        endpoint: 'jsonp',
-        execCallback: 'Jazzcat.combo.exec',
-        loadCallback: 'Jazzcat.combo.load',
-        projectName: ''
-    };
+            if (jsonp) {
+                // Insert the httpCache loader before the first script
+                if (i===0) {
```
noahadams: What if the first script is inline or fails the httpUrl test? We can't just check the index for this, you'll need a variable to tell you if you've seen the first "eligible" script.

noahadams: could also filter the list of scripts for elligibility before we get to this loop

build/mobify-2.0.0alpha4.js
```diff
@@ -1626,14 +1631,15 @@ define('jazzcat',["utils", "capture"], function(Utils, Capture) {
...
      *   After:
      *
-     *   <script>true,"body",Jazzcat.combo.exec("http://code.jquery.com/jquery.js")</script>
+     *   <script>Jazzcat.exec("http://code.jquery.com/jquery.js")</script>
      *   <script>$(function() { alert("helo joe"); })</script>
      *
-     * Note that this only the first part of the Jazzcat transformation. The
-     * bootloader script is inserted by the overriden `Capture.enabled` function.
+     * Note that this is only the first part of the Jazzcat transformation. The
+     * bootloader script is inserted by the overriden `Capture.enabled`
```
noahadams: plz update this comment

noahadams:
:+1:

jansepar:
Thanks @noahadams!

PR #186, opened 2013-07-29
Fixes a regression that will cause inline scripts to be removed when using js responseType

jansepar:


noahadams:
this looks good. :+1:

PR #189, opened 2013-07-30
adds a rather barebones timing api and some tests

noahadams:


rrjamie:
this looks good to me. @jansepar can you take a quick look?

Gruntfile.js
```diff
@@ -116,7 +136,8 @@ module.exports = function(grunt) {
                         'http://localhost:3000/tests/resizeImages.html',
                         'http://localhost:3000/tests/jazzcat.html',
                         'http://localhost:3000/tests/unblockify.html',
-                        'http://localhost:3000/tests/cssOptimize.html'
+                        'http://localhost:3000/tests/cssOptimize.html',
+                        'httpp://localhost:3000/tests/timing.js'
```
jansepar: extra p

jansepar:
While I do feel that creating a new library that includes the timing API is the wrong approach, I think it's fine for now. Eventually we should just be creating a custom mobify.js bundle for Swift that also requires in the swift processor stuff (which isn't that far away).

noahadams:
@jansepar I basically agree, but this is the quick and dirty approach that will work most easily for us now with a  minimum of consequence.

jansepar:
Yep agreed :) I'm going to make sure everything is still working and run the tests on Sauce before I give my +1

jansepar:
:+1:

PR #194, opened 2013-09-10
V2.0 streaming capturing

jansepar:
DO NOT MERGE!! Open for review.

TODO:
- ~~Insert library back into the captured document~~
- ~~Move meta tags into main document~~
- ~~Move title tag into main document~~
- ~~Move HTML/HEAD attributes into HTML/HEAD tags in iframe~~
- ~~Solve referer issue~~
- ~~Fix window.location.~~
- ~~Pass a capture object instead of a captured doc to the chunk callback~~
- ~~Fix iframe issue on Android 2.3~~
- ~~Fix Jazzcat tests~~
- Write amazing tests


src/capture.js
```diff
@@ -86,29 +86,68 @@ function extractHTMLStringFromElement(container) {
     }).join('');
 }
 
+/**
+ * Takes a method on one source object and overrides it to call the method on
+ * a destination object (in addition to calling the method on the source object)
+ */
+var applyMethodToDifferentObject = function(srcObj, destObj, method) {
```
noahadams: maybe a name like bindToAddtionalObject or invokeWithSameNameOnOtherObject? it's not clear from the comments that the "method" argument needs to be a name, rather than the method itself.

src/capture.js
```diff
@@ -86,29 +86,68 @@ function extractHTMLStringFromElement(container) {
...
+ * Removes all closing tags from an html string
+ */
+var removeCloseEndTags = function(html) {
+    var newHtml = html.replace(/(.*)<\/[^>]*>$/i, function(match, p1){
+        return p1
+    });
+    if (newHtml === html) {
+        return newHtml;
+    }
+    return removeCloseEndTags(newHtml);
```
noahadams: could we make this iterative rather than recursive? i.e. 

```
while(newHtml !== html) {
    html = newHtml;
    newHtml = html.replace(/(.*)<\/[^>]*>$/i, function(match, p1){ return p1 }); 
}
```

src/capture.js
```diff
@@ -86,29 +86,68 @@ function extractHTMLStringFromElement(container) {
...
-
-    var capturedStringFragments = this.createDocumentFragmentsStrings();
-    Utils.extend(this, capturedStringFragments);
-
-    var capturedDOMFragments = this.createDocumentFragments();
-    Utils.extend(this, capturedDOMFragments);
 };
 
-var init = Capture.init = function(callback, doc, prefix) {
+Capture.init = Capture.initCapture = function(callback, doc, prefix) {
```
noahadams, line 242: Why does this get two names?

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
 };
 
 /**
+ * Streaming capturing is a batshit loco insane way of being able to modify
```
noahadams: batcrap, this is polite open source here

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+    var main = Capture.getMain();
+    if (main) {
+        startCapturedHtml += Utils.outerHTML(main);
+    }
+
+    if (iframe) {
+        // In Webkit/Blink, resources requested in a non-src iframe do not have
+        // a referer attached. This is an issue for scripts like Typekit.
+        // We get around this by manipulating the browsers
+        // history to trick it into thinking it is an src iframe.
```
noahadams: into sending a referer

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+    // Start the captured doc with the original pieces of the source doc
+    var startCapturedHtml = Capture.getDoctype(sourceDoc) +
+                 Capture.openTag(sourceDoc.documentElement) +
+                 Capture.openTag(sourceDoc.head) +
+                 // Even if there is another base tag in the site that sets
+                 // target, the first one declared will be used
+                 // TODO: Write tests to verify this for all of our browsers.
+                 '<base target="_parent" />' +
+                 // Grab and insert all existing HTML above plaintext tag
+                 extractHTMLStringFromElement(sourceDoc.head);
```
noahadams, line 353: this should perhaps be unit tested

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+
+    if (iframe) {
+        // In Webkit/Blink, resources requested in a non-src iframe do not have
+        // a referer attached. This is an issue for scripts like Typekit.
+        // We get around this by manipulating the browsers
+        // history to trick it into thinking it is an src iframe.
+        // AKA an insane hack for an insane hack.
+        try {
+            iframe.contentWindow.history.replaceState({}, iframe.contentDocument.title, window.location.href);
+        } catch (e) {} // This will throw an exception on Firefox
```
noahadams: This is fishy

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+    startCapturedHtml = Capture.disable(startCapturedHtml, prefix);
+
+    // Start the captured doc off write! (pun intended)
+    capturedDoc.write(startCapturedHtml);
+
+    // Track what has been written to captured and destination docs for each chunk
+    var plaintextBuffer = '';
+    var writtenToDestDoc = '';
+
+    var pollPlaintext = function(){
```
noahadams: should probably pull this up a level and make it take its closed over variables as arguments.

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+        if (toWrite === '' && !finished) {
+            setTimeout(pollPlaintext, pollInterval);
+            return;
+        }
+
+        // Write our progress to plaintext buffer
+        plaintextBuffer += toWrite;
+
+        // Escape resources for chunk and remove target=self
+        toWrite = Capture.disable(toWrite, prefix).replace(/target="_self"/gi, '');
```
noahadams: html attrs might also be quoted with single quotes, or not quoted at all, this operation might be good to split off into a function and test, or manipulate elements on their way "out" of the intermediate DOM iframe (since we expect no navigation to happen in there). It's also possible for the target attributes of links to get set programmatically, or to come in as the result of AJAX calls after we're done fiddling with the document. (what I'm saying is we might need to hook into navigation at a lower level).

Seems fishy to just chain it on here

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+
+        // Write escaped chunk to captured document
+        capturedDoc.write(toWrite);
+
+        if (iframe) {
```
noahadams: this variable could better named, I've definitely forgotten what it is by this point

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+                }
+            }
+
+            // In Android 2.3, widths of iframes and cause an override of the width
+            // if the html element of the top-level document. We detect for that
+            // and change the width of the iframe
+            if (document.documentElement.offsetWidth !== window.outerWidth) {
+                var iframes = Array.prototype.slice.call(capturedDoc.querySelectorAll('iframe'));
+                //Utils.removeElements(capturedDoc.querySelectorAll('iframe'));
+                iframes.filter(function(iframe){
```
noahadams: filter is definitely not a natural iteration function, maybe a foreach or a for loop?

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+                    el.setAttribute('capture-moved', '');
+                }
+            }
+
+            // In Android 2.3, widths of iframes and cause an override of the width
+            // if the html element of the top-level document. We detect for that
+            // and change the width of the iframe
+            if (document.documentElement.offsetWidth !== window.outerWidth) {
+                var iframes = Array.prototype.slice.call(capturedDoc.querySelectorAll('iframe'));
+                //Utils.removeElements(capturedDoc.querySelectorAll('iframe'));
```
noahadams: this should go

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+                    if (el.hasAttribute('capture-moved')) {
+                        continue;
+                    }
+                    var elClone = sourceDoc.importNode(el, true);
+                    sourceDoc.head.appendChild(elClone);    
+                    el.setAttribute('capture-moved', '');
+                }
+            }
+
+            // In Android 2.3, widths of iframes and cause an override of the width
```
noahadams: more english

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+            // and change the width of the iframe
+            if (document.documentElement.offsetWidth !== window.outerWidth) {
+                var iframes = Array.prototype.slice.call(capturedDoc.querySelectorAll('iframe'));
+                //Utils.removeElements(capturedDoc.querySelectorAll('iframe'));
+                iframes.filter(function(iframe){
+                    iframe.width = '100%';
+                });
+            }
+            // var i = capturedDoc.createElement('iframe');
+            // capturedDoc.body.appendChild(i);
```
noahadams: pew pew pew

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+
+        if (capturedDoc.documentElement) {
+            // Grab outerHTML of capturedDoc and write the diff to destDoc
+            capturedHtml = Utils.outerHTML(capturedDoc.documentElement);
+            // we could be grabbing from a captured document that has a head and no body.
+            var toWriteDest = capturedHtml.substring(writtenToDestDoc.length);
+
+            // outerHTML will always give us an balanced tree, which isn't what
+            // we want to write into the destination document. The solution for
+            // this is to simply never write out closing head tags if they
```
noahadams: s/closing head tags/closing tags/

src/capture.js
```diff
@@ -133,6 +172,237 @@ var init = Capture.init = function(callback, doc, prefix) {
...
+            // are at the end of the `toWriteDest` string. If those end tags
+            // were truly from the document, rather then generated by outerHTML,
+            // then they will come in on the next chunk.
```
noahadams: (or be included with the last chunk)

src/capture.js
```diff
@@ -329,9 +600,6 @@ Capture.prototype.restore = function() {
  * Set the content of an element with html from a string
  */
 Capture.prototype.setElementContentFromString = function(el, htmlString) {
-    // We must pass in document because elements created for dom insertion must be
-    // inserted into the same dom they are created by.
-    var doc = this.doc;
     for (cachedDiv.innerHTML = htmlString; cachedDiv.firstChild; el.appendChild(cachedDiv.firstChild));
```
noahadams, line 615: should we sanitize cached div here? (eg cachedDiv.innerHTML = '') does anything else use it? perhaps it should be declared down here.

src/capture.js
```diff
@@ -413,39 +681,60 @@ Capture.prototype.getCapturedDoc = function(options) {
...
+
+    return mobifyjsScript;
+};
+
+Capture.getMain = function(doc) {
+    var doc = doc || document;
+    var mainScript = undefined;
+    if (window.Mobify && window.Mobify.mainExecutable) {
+        mainScript = document.createElement('script');
+        mainScript.innerHTML = "var main = " + window.Mobify.mainExecutable + "; main();";
```
noahadams: an explicit toString() of the function and some documenting comments would be welcome here.

src/jazzcat.js
```diff
@@ -309,13 +311,20 @@ define(["utils", "capture"], function(Utils, Capture) {
             return scripts;
         }
 
+        // Remove mobify script(s) from scripts array
```
noahadams: this seems a little too specific (it assumes 1 or two mobify scripts at the start). Maybe here a for-loop or .filter() might be appropriate, could even throw this logic into the filtering for-loop below.

src/jazzcat.js
```diff
@@ -337,22 +346,32 @@ define(["utils", "capture"], function(Utils, Capture) {
...
-            var httpLoaderScript = Jazzcat.getHttpCacheLoaderScript();
-            scripts[0].parentNode.insertBefore(httpLoaderScript, scripts[0]);
-        }
 
         for (var i=0, len=scripts.length; i<len; i++) {
             var script = scripts[i];
 
-            url = script.getAttribute(options.attribute);
+            // Skip script if it has been optimized already
+            if (script.hasAttribute('mobify-optimized') || script.hasAttribute('skip-optimize')) {
```
noahadams: should probably document that "skip-optimize" attribute, also, should we be adding "data-" attributes?

src/jazzcat.js
```diff
@@ -337,22 +346,32 @@ define(["utils", "capture"], function(Utils, Capture) {
...
+                continue;
+            }
 
             // skip if the script is inline
+            url = script.getAttribute(options.attribute);
             if (!url) continue;
```
noahadams: style: this if should be a suiggly brace block as above

src/jazzcat.js
```diff
@@ -337,22 +346,32 @@ define(["utils", "capture"], function(Utils, Capture) {
...
             url = Utils.absolutify(url);
             if (!Utils.httpUrl(url)) continue;
```
noahadams: ditto ^^

src/jazzcat.js
```diff
@@ -337,22 +346,32 @@ define(["utils", "capture"], function(Utils, Capture) {
...
 
+            // TODO: Check for async/defer
+
+            // Load what we have in http cache, and insert loader into document
+            if (jsonp && !Jazzcat.loaderInserted) {
+                httpCache.load(httpCache.options);
+                var httpLoaderScript = Jazzcat.getHttpCacheLoaderScript();
+                script.parentNode.insertBefore(httpLoaderScript, script);
+                // ensure this doesn't happen again for this page load
+                Jazzcat.loaderInserted = true;
```
noahadams: Let's call this `Jazzcat.cacheLoaderInserted` for clarity

src/resizeImages.js
```diff
@@ -279,12 +240,14 @@ ResizeImages.resize = function(elements, options) {
...
         // For an `img`, simply modify the src attribute
-        if (element.nodeName === 'IMG') {
+        if (element.nodeName === 'IMG' && !element.hasAttribute('mobify-optimized')) {
+            element.setAttribute('mobify-optimized', '');
             ResizeImages._rewriteSrcAttribute(element, opts);
         }
         // For a `picture`, (potentially) nuke src on `img`, and
         // pass all `source` elements into modifyImages recursively
-        else if (element.nodeName === 'PICTURE') {
+        else if (element.nodeName === 'PICTURE' && !element.hasAttribute('mobify-optimized')) {
```
noahadams: as discussed, we need to do this on `<source>` elements as we might find ourselves running this on a fragment of a picture element whose other fragment has not yet been streamed

src/utils.js
```diff
@@ -67,11 +67,16 @@ Utils.httpUrl = function(url) {
  * outerHTML polyfill - https://gist.github.com/889005
  */
 Utils.outerHTML = function(el){
-    var div = document.createElement('div');
-    div.appendChild(el.cloneNode(true));
-    var contents = div.innerHTML;
-    div = null;
-    return contents;
+    if (el.outerHTML) {
```
noahadams, line 70: ! glad we have this now

src/warmOnTap.js
```diff
@@ -4,7 +4,7 @@
 // Developer can pass an optional parameter specifying selector match for elements to trigger on
 // Defaults to some common targets.
 
-define(["warmOnTap"], function(warmOnTap) {
+define(["mobifyjs/warmOnTap"], function(warmOnTap) {
```
noahadams: is it just me or does this module require itself?

tag/bootstrap.html
```diff
@@ -15,14 +15,9 @@
     return;
 }
 
-function attachScriptBeforeMobifyTag(script, id, classNames, data) {
+function attachScriptBeforeMobifyTag(script, id, classNames, src) {
```
noahadams, line 18: what's motivating these tag changes?

src/capture.js
jansepar, line 615: What do you mean by sanitize it? Like set it to blank afterwards? Also its declared elsewhere because it's used by multiple functions, to avoid needing to recreated a div each time.

src/warmOnTap.js
jansepar: lol good catch - never noticed that. I have never used this script :)

tag/bootstrap.html
jansepar, line 18: This tag used to take a "main" function and converted into an inline script, both so it could execute and for it to be moved over the flood. But inserting the script dynamically like that did not work for streaming capturing in Firefox - so instead, I just execute the function (if it's a function), and leave the work of creating a script node out of it to capture.js.

src/utils.js
jansepar, line 70: Me too :)

src/jazzcat.js
jansepar: Yep definitely - once we write up Jazzcat docs I will document that for sure.

noahadams:
:+1: let's make sure we add more test coverage next week

PR #195, opened 2013-09-18
Removing broken Twitter links

mobifydave:


mobifydave:
Based the PR on master not v2, so it won't have an effect.

PR #197, opened 2013-10-03
Pre rendering bug in iframe for iOS 6 and 7, and Android 2.3

jansepar:
Adds a fix for width of the document element getting too large because of a `pre` rendering bug in iframes for iOS6/7 and Chrome 2.3


kbingman:
Ok, this looks good. :+1:

jansepar:
Thanks @kbingman!

PR #198, opened 2013-10-03
Fix for Jazzcat to handle scripts that have onload handlers attached

jansepar:


src/jazzcat.js
```diff
@@ -388,8 +388,14 @@ define(["mobifyjs/utils", "mobifyjs/capture"], function(Utils, Capture) {
                 }
                 script.type = 'text/mobify-script';
                 // Rewriting script to grab contents from our in-memory cache
-                // ex. <script>Jazzcat.combo.exec("http://code.jquery.com/jquery.js")</script>                    
-                script.innerHTML =  options.execCallback + "('" + url + "');";
+                // ex. <script>Jazzcat.exec("http://code.jquery.com/jquery.js")</script>
+                if (script.hasAttribute('onload')){
+                    var onload = script.getAttribute('onload');
+                    script.innerHTML =  options.execCallback + "('" + url + "', '" + onload.replace(/'/g, '\\\'') + "');";
```
noahadams, line 394: What's with the single quote escaping of the onload handler?

jansepar, line 394: Since we're wrapping the value with single quotes, we need to ensure that what is in the onload string doesn't contain single quotes. We're constructing a string version of the exec call.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mobifyjs",
-  "version": "2.0.0alpha7",
+  "version": "2.0.0alpha9",
```
jansepar, line 3: Just a heads up, this is because for my last tag release, I forgot to bump the version, thus the increase by 2.

noahadams:
@jansepar I did a little code cleanup on the file and added to the example, if you're cool with those you have my :+1:

jansepar:
Your changes look good to me! Thanks @noahadams

PR #204, opened 2013-10-11
Use orientationchange instead of resize for adjusting widths on certain browsers with streaming capturing

jansepar:


src/capture.js
```diff
@@ -350,7 +350,7 @@ Capture.initStreamingCapture = function(chunkCallback, finishedCallback, options
     // to grow larger because the width of the `pre/iframe`.
     if (/ip(hone|od|ad)|android\s2\./i.test(navigator.userAgent)) {
         explicitlySetWidth();
-        window.onresize = function(){
+        window.onorientationchange = function(){
```
rrjamie: You might want to use `addEventListener`, in case they use this and your handler gets obliterated... though we are in the outside iframe so maybe it's okay.

jansepar: Ya I figured they were using different windows. But, I've changed it just incase. Have another look!

build/mobify.js
```diff
@@ -989,13 +989,17 @@ Capture.initStreamingCapture = function(chunkCallback, finishedCallback, options
...
+    var match = /ip(hone|od|ad)|android\s2\./i.exec(navigator.userAgent);
+    if (match) {
         explicitlySetWidth();
-        window.onresize = function(){
-           setTimeout(function(){
-               explicitlySetWidth();
-           }, 0);
-        }
+        var ios = (match[1] !== undefined);
+        debugger;
```
rrjamie: ahem.

rrjamie:
tested on android 2.3.5, 2.3.6, 4.0.1 and ios 7.0.2

works

jansepar:
@rrjamie thanks!

PR #205, opened 2013-10-16
V2.0 streaming height fix

jansepar:
- Fix for incorrectly sized scroll bars in non-iOS Safari browsers.
- Removing explicit width setting on iframe that seems to be causing iOS7 Safari to crash on certain sites.


jansepar:
It was determined that the height fix in this PR caused issues with position: fixed. Issuing new PR that carries over the crash fix - will figure out an issue for the heights at a later date.

PR #206, opened 2013-10-19
V2.0 bug fixes

jansepar:
- Fixes an issue where setting the width of the iframe on rotation causes Safari on iOS to crash.
- Forces the loading bar to remain until the CSS completes downloading. This is done by loading the CSS in the main document, because resources loaded through document.write do not initiate the loading bar.
- Expands the height of an iframe to the contents within.


src/capture.js
```diff
@@ -394,6 +401,37 @@ Capture.initStreamingCapture = function(chunkCallback, finishedCallback, options
...
+                 var heightElement = webkit ? capture.destDoc.documentElement : capture.destDoc.body;
+                 if (capture.destDoc.documentElement === null || capture.destDoc.body === null) {
+                     return;
+                 }
+                 // Sometimes, documentElement can have a scroll height of 0.
+                 // If so, set the height of it to 100% and attempt to get it again.
+                 var height = heightElement.scrollHeight;
+                 if (height === 0) {
+                    //capture.destDoc.documentElement.style.height = '100%';
+                    //capture.destDoc.body.style.height = '100%';
```
kbingman, line 421: @jansepar shouldn't we clear these comments out?

kbingman:
so other that the weird height issue for that one site this looks good. :+1:

PR #208, opened 2013-10-22
Guard hiding when re-inserting user CSS into page

jansepar:


jansepar:
Found a much better way to do this. Closing branch.

PR #211, opened 2013-10-25
Support IE10+ instead of just IE10, added test for IE11

philklc:


jansepar:
Awesome thanks for the PR! I should be able to get to it sometime this weekend :)

jansepar:
:+1: thanks again @philklc!

PR #215, opened 2013-10-31
image size binning

noahadams:
Modifies image resizer defautls in MobifyJS 2.0:

If not otherwise set, images' width bounds will be restricted to values of 320, 640, 720, 1024, 1536, 2048 or 4000 pixels, whichever is closest and greater than or equal to the device's current width (i.e. this might change with orientation).


noahadams:
~~I'd mention that this PR is pretty opinionated: It resizes things by creating a bounding box that is a square with dimensions equal to the longest pixel dimension of the device. this ensures that we provide a good size for both portrait and landscape orientation if the device is rotated while viewing the page.~~

~~I realize "delivering more pixels" might be controversial, but binning by necessity will be sending more pixels than stricter sizing, and using square bounding boxes allows us to have fewer "bins".~~

~~I want to prepare a bit of a demo around size vs. compression ratio to aid in decision making around this.~~

Works as described above. File size demos went badly but opened up some interesting new cans of worms.

src/resizeImages.js
```diff
@@ -201,13 +199,31 @@ ResizeImages._crawlPictureElement = function(el, opts) {
...
+ * the passed value and return it, or return the greatst value if none are 
+ * greater
+ */
+var targetDims = [320, 640, 720, 1024, 1536, 2048, 4000];
+ResizeImages._getBinnedDimension = function(dim) {
+    var resultDim = 0;
+
+    for(var i = 0, len = targetDims.length; i < len; i++) {
+        resultDim = targetDims[i];
+        if(resultDim >= dim) {
```
rrjamie: `if<space>(`, `for<space>(`)

src/resizeImages.js
```diff
@@ -216,10 +232,10 @@ ResizeImages.resize = function(elements, options) {
 
     var screenSize = Utils.getPhysicalScreenSize(dpr);
 
-    // If maxHeight/maxWidth are not specified, use screen dimentions
+    // If maxHeight/maxWidth are not specified, use screen dimensions
     // in device pixels
-    var width = opts.maxWidth || screenSize.width;
-    var height = opts.maxHeight || screenSize.height;
+    var width = opts.maxWidth || ResizeImages._getBinnedDimension(screenSize.width);
+    var height = opts.maxHeight || undefined;
```
rrjamie, line 238: Whats the reasoning behind no longer providing a height by default?

noahadams, line 238: My motivation is to decrease the potential number of bins by removing a dimension from consideration. Shawn also commented that he was unsure why we were restricting the height in the first place and he believed we should be restricting widths.

src/resizeImages.js
noahadams: fair

noahadams:
Made changes, added built files, doing a fresh master merge now.

src/resizeImages.js
```diff
@@ -201,13 +199,31 @@ ResizeImages._crawlPictureElement = function(el, opts) {
 };
 
 /**
+ * Searches a list of target dimensions for the smallest one that is greater than 
+ * the passed value and return it, or return the greatst value if none are 
+ * greater
+ */
+var targetDims = [320, 640, 720, 1024, 1536, 2048, 4000];
```
rrjamie: I don't think these choices of bins make sense. Here are some sizes of common/popular/major phones:

iPhone (Non-Retina)     320     480  
iPhone 4                640     960         30%~
iPhone 5                640     1136        30%~
Galaxy S3               720     1280        10%
Nexus  4                768     1280
Nexus  5                1080    1920
HTC One                 1080    1920
LG G2                   1080    1920
iPad (Non-Retina)       768     1024  
iPad (Retina)           1536    2048
Lumina Series           480     800
Lumina Series           768     1280
Lumina Series           1080    1920

It looks like the sizes we should target are: 320 (iPhone, Shitty Phones), 640 (iPhone Retina), 768 (720p and iPad 2), 1080 (1024px and 1080p Phones), 1536 (iPad Retina), then 2048 (Who knows), and 4000 (Max).

noahadams: 2048 is retina ipad in landscape (and a reasonable catcher for 1920).

Your bin suggestions seem reasonable, would cause a few things not to get promoted as far...

noahadams: doneskies

rrjamie:
+1

build/mobify.js
```diff
@@ -1693,13 +1691,31 @@ ResizeImages._crawlPictureElement = function(el, opts) {
 };
 
 /**
+ * Searches a list of target dimensions for the smallest one that is greater than 
+ * the passed value and return it, or return the greatst value if none are 
+ * greater
+ */
+var targetDims = [320, 640, 768, 1080, 1536, 2048, 4000];
```
jansepar, line 1698: It would be great to get a comment here (basically @rrjamie's comment about the common devices) so that people taking a peek into this code know why we make this decision.

PR #217, opened 2013-11-06
<img> onerror investigation

noahadams:
Adds an example that illustrates when an "error" event is triggered on image elements (i.e. when we expect our onerror code in the image resizer client code to be triggered).


jansepar:
:+1:

PR #218, opened 2013-11-06
Copy updates to site home page

mobifydave:
Linking to MozHack's capturing article, reflowing intro copy, and updating copy at the bottom of the page to talk about MPS.


jansepar:
:+1:

PR #224, opened 2013-11-07
Bump version to 2.0.0 stable!

jansepar:


noahadams:
Oh my. We're finally here. and with so little fanfare. Watch out for version 2.2 (and 2.0.1)! :+1:

PR #226, opened 2013-11-13
Changes getImageURL to grab default options if options object is not passed to it

jansepar:


build/mobify.js
```diff
@@ -1420,6 +1416,17 @@ ResizeImages.resize = function(elements, options) {
...
+    return opts;
+}
+
+/**
+ * Searches the collection for image elements and modifies them to use
+ * the Image Resize service. Pass `options` to modify how the images are 
+ * resized.
+ */
+ResizeImages.resize = function(elements, options) {
+    var opts = ResizeImages.getDefaultOptions(options);
```
noahadams: This changes the interface to this function. Previously width and height parameters on the options object were expressed in css pixels, now they must be expressed in device pixels to produce the same url.

noahadams:
@jansepar, I went in and did a little renaming, with it this gets :+1: from me ;)

PR #234, opened 2013-12-10
PROD-1070 - Fixes Anchor Tags on Firefox

rrjamie:
**Status**: Review
**Review**: @jansepar 
**JIRA**: [PROD-1070](https://mobify.atlassian.net/browse/PROD-1070), [PROD-1155](https://mobify.atlassian.net/browse/PROD-1155)
## Changes
- (On Firefox) Adds an event handler to every link that properly handles anchor tags, and patches up `preventDefault` anchor scrolling can be cancelled by the usual methods (returning `false`, calling `preventDefault`, etc).
- Adds tests (for all browsers) to test that anchor tags behave properly
## How to test-drive this PR
- Run the tests
- If you need you to manually test it, visit http://localhost:3000/tests/fixtures/anchor-test.html and tap the links.
## Notes

On Android 4.0.x, iFrames don't behave well with scrolling and anchor tags. However, manually testing shows that it works on 4.0.x.
## Todo
- Test on Android 2.x
- I'm not sure where we should put this patching code. There isn't an obvious place to put it so it runs after capturing, without requiring the user to explicitly add the patching code themselves.


build/mobify.js
```diff
@@ -2200,6 +2231,122 @@ var defaults = CssOptimize._defaults = {
...
+            }
+            
+            // Newer browsers support `e.defaultPrevented`. FF 4.0 supports `e.getPreventDefault()`
+            var defaultPrevented = (typeof e.defaultPrevented !== "undefined") ?
+                e.defaultPrevented :
+                e.getPreventDefault && e.getPreventDefault();
+
+            if (!defaultPrevented) {
+                // Prevent the default action, which would cause a
+                // page relresh.
```
noahadams: refresh

mobify-custom.js.example
```diff
@@ -29,6 +29,8 @@ require(["mobifyjs/utils", "mobifyjs/capture", "mobifyjs/resizeImages", "mobifyj
 
             capture.renderCapturedDoc();
         });
+    } else {
+        FF.patchAll();
```
jansepar: I don't think this is something the user should have to execute to fix broken FF. I feel like since Capturing is what is causing the breakage, it should be Capturing's responsibility to fix. I would envision us running the "patchAll()" method in Capture.js itself when not in "capturing" mode.

src/firefox.js
```diff
@@ -0,0 +1,115 @@
+// Fixes specific to Firefox.
+
+define(["mobifyjs/utils"], function(Utils){
```
jansepar: Kind of similar to my previous comment, but I'm not sure that it makes sense to have a "firefox.js" here. I think this should be part of capture.js.

rrjamie: Well, I think all FF specific code should live on it's own somewhere, not cluttering up capturing.js

mobify-custom.js.example
rrjamie: Well, it shouldn't be, but there is currently no place where we handle what happens post-flood. We only have the `if(Mobify.capturing) {} else {}` block. Do you have a good idea where you would want to call this from?

src/firefox.js
jansepar: Hmmm. I think then the best solution is to keep the FF code separate, but make it a dependancy of capturing.js instead of mobify-library.js. The way I'm thinking about it is, imagine capturing.js was a stand-alone library - it would still need the FF fixes.

rrjamie: still, the question is "where is the entry point" for code that should be called after capturing, cause afaik there isn't one yet.

mobify-custom.js.example
jansepar: One thing I was thinking was that we could stick `if(Mobify.capturing) {} else {FF.patchAll()}` in Capturing itself - but the lame part about that is if someone included Mobify.js in their site, but didn't use Capturing, then the FF patch would be unnecessarily applied.

I think the "best" solution solution here is to inject an inline script into the capturedDoc that calls FF.patch() so that it executes post-flood. I'd also prefer that we only inject it if we see FF, instead preventing it from doing anything if its not FF in the function itself.

src/firefox.js
jansepar: Just responded to your other comment for that answer ;)

mobify-custom.js.example
rrjamie: Well, I'd argue then we always add a `<script>Mobify.postCapture()</script>` to all pages and then we have a centralized place to put all this stuff

jansepar: Yeah sure I'm cool with that.

rrjamie:
@jansepar here you go

build/mobify.js
```diff
@@ -634,12 +634,159 @@ Utils.getPhysicalScreenSize = function(devicePixelRatio) {
...
+            }
+
+            // Scroll to it, if it exists
+            if (target) {
+                target.scrollIntoView && target.scrollIntoView();
+            }
+        }
+
+        // We have to get the event through bubbling, otherwise
+        // events cancelled by a the return value of an onclick
```
jansepar: typo?

build/mobify.js
```diff
@@ -634,12 +634,156 @@ Utils.getPhysicalScreenSize = function(devicePixelRatio) {
...
+    // Backup with polling incase readystatechange doesn't fire
+    // (happens with some Android 2.3 browsers)
+    var iid = setInterval(function(){
+        if (Utils.domIsReady(doc)) {
+            onReady();
+        }
+    }, 100);
+
+    doc.addEventListener("readystatechange", onReady, false);
+};
```
jansepar, line 668: It would be nice to use this method in Capturing to DRY things up. But thats a more dangerous change that I'd rather not do now :)

jansepar:
:+1: great job on these tests!

PR #235, opened 2013-12-12
Adds a hotfix to ensure that meta tags do not get created and inserted into the <head> for older Safari browsers

jansepar:
**Status**: Review
**Review**: @rrjamie  
## Changes
- Fixes an issue where in Safari 4/5 and iOS 4.3, there are certain scenarios where elements in the body (ex "meta" in "noscripts" tags) get moved into the head, which can cause issues with certain websites (for example, if you have a meta refresh tag inside of a noscript tag)
- Adds regression test
## How to test-drive this PR
- Download Safari 4.0.5 (http://mirror.jquery.com/multi-safari/multi-safari-4.0.5-snow-leopard.zip)
- Run the capture.html test on Safari 4.0.5, on Chrome, on Phantomjs (`grunt test`) and on Saucelabs (`grunt saucelabs`)


build/mobify.js
```diff
@@ -1001,10 +1000,21 @@ Capture.prototype.createDocumentFragments = function() {
...
-    var disabledHeadContent = Capture.disable(this.headContent, this.prefix);
+
+    // In Safari 4/5 and iOS 4.3, there are certain scenarios where elements
+    // in the body (ex "meta" in "noscripts" tags) get moved into the head,
+    // which can cause issues with certain websites (for example, if you have
+    // a meta refresh tag inside of a noscript tag)
+    var heads = doc.querySelectorAll('head');
+    if (heads.length > 1) {
+        while (heads[1].hasChildNodes()) {
+            heads[1].removeChild(heads[1].lastChild);
```
noahadams, line 1011: this is like one of my least favourite special case patches, but i guess it does only modify things in `heads[1]` which shouldn't exist to begin with...

noahadams:
:+1:

jansepar:
@noahadams thanks!

PR #236, opened 2013-12-17
Fix bugs with the mobify tag being placed above the opening head tag in order browsers

jansepar:
**Status**: Review
**Review**: @noahadams 
## Changes
- In older browsers (Safari 4/iOS 4.3), placing the tag above opening `<head>` causes issues. To fix this, we must properly grab the `headContent` when splitting up the plaintext string.
## How to test-drive this PR
- Download Safari 4.0.5 (http://mirror.jquery.com/multi-safari/multi-safari-4.0.5-snow-leopard.zip)
- Run the capture.html test on Safari 4.0.5, on Chrome, on Phantomjs (`grunt test`) and on Saucelabs (`grunt saucelabs`), and Android 2.3


src/capture.js
```diff
@@ -282,16 +289,26 @@ Capture.openTag = function(element) {
         // <!-- comment --> . Skip it.
         if (!match[1]) continue;
 
+        // Grab the contents of head
+        captured.headContent = rawHTML.slice(0, match.index);
+
+        // Parse the head content
+        var parsedHeadTag = /^\s*(<head[^>]*>)([\s\S]*)/gi.exec(captured.headContent);
```
noahadams: why is the second matching group not ".*"?

tests/capture.html
```diff
@@ -184,6 +188,71 @@
...
+            bodyOpenTag: "<body bar=\"baz\">",
+            headContent: "\n    \n    <link rel=\"stylesheet\" href=\"/path/to/stylesheet.css\">\n</head>\n",
+            bodyContent: "\n    <p>Plaintext example page!</p>\n    <script src=\"/path/to/script.js\"><\/script>\n</body>\n</html>\n\n"
+        }
+
+        asyncTest("createDocumentFragmentsStrings - below head tag", function(){
+            var iframe = $("<iframe>", {id: "plaintext-example"});
+            iframe.attr("src", "/tests/fixtures/plaintext-example.html")
+            iframe.load(function(){
+                var doc = this.contentDocument;
```
noahadams, line 205: just wondering what inspired you to use `this` in the load handler rather than the `iframe` variable int eh enclosing scope?

noahadams:
tests look good, answer my questions, get my approval!

src/capture.js
jansepar: "." does not capture line break characters in JS. Other languages allow you to configure what "." captures, but JS doesn't. The standard way around this is by using the \s\S convention.

http://www.regular-expressions.info/dot.html

tests/capture.html
jansepar, line 205: Not sure tbqh :). I just copy and pasted a previous test with this iframe logic - I wrote the original a while back. I'm sure either would work fine.

jansepar, line 205: I just went to switch it to `iframe`, but realized that it's a jquery object, meaning I'd have to replace `this` with `iframe[0]` or `iframe.get(0)`, which I think is worse.

noahadams:
:+1:

tests/capture.html
johnboxall, line 205: @jansepar to avoid this sort of thing, I typically prefix jQuery variables with `$` eg. `$iframe`!

build/mobify.js
```diff
@@ -923,16 +929,27 @@ Capture.openTag = function(element) {
         // <!-- comment --> . Skip it.
         if (!match[1]) continue;
 
+        // Grab the contents of head
+        captured.headContent = rawHTML.slice(0, match.index);
+        // Parse the head content
+        // (using a "new RegExp" here because in Android 2.3 when you use a global
+        // match using a RegExp literal, the state is incorrectly cached).
+        var parsedHeadTag = (new RegExp('^\\s*(<head(?:[^>\'"]*|\'[^\']*?\'|"[^"]*?")*>)([\\s\\S]*)$')).exec(captured.headContent);
```
rrjamie, line 937: I usually just avoid `exec` because of it's weird behavior and use `string.match(re)` not sure if still borked on Android 2.3

tests/capture.html
```diff
@@ -47,6 +47,10 @@
     QUnit.config.autostart = false;
     QUnit.config.testTimeout = 30 * 1000;
 
+    var removeNewlines = function(str) {
+        return str.replace(/(\r\n|\n|\r)/gm,'');
```
rrjamie, line 51: I think you can just go `(\r|\n)`

build/mobify.js
jansepar, line 937: Yeah I switched it up to match at one point, which did work as well.

jansepar:
:+1:

PR #237, opened 2013-12-19
New Tag - Initial Work

rrjamie:
New Tag

This is some initial work on the new tag. This is a rewrite from the start, however, I haven't tested it with any builds. I'm trying to get a nice clean interface, and a testable setup.

I see three stages to the new tag:

1) Write a new tag, and review if we like the interface.  (This Stage)

I really want feedback here on the interface, since I moved away from the "detector" function.

2) Make it compatible with MPS/1.1.

3) Make it small.

Feedback on (2) and (3) is cool too, but only really if decisions in (1) make stuff really difficult in (2) and (3).

**Status**: _Review_

**Reviewers**: @jansepar (optionally @noahadams @tedtate)
**JIRA**: PROD-778, PROD-1040
## Changes
- Completely new tag, written in parallel with tests.
- Includes all features (I think)
  - Capturing based on device (or mode)
  - Loading a script based on device (or mode)
  - Loading preview
  - Unmobify via Reload
  ## \- Made to be testable.
- Not yet tested with actual builds, looking for feedback on design and interface (the `options` object)
## How to test-drive this PR
- There are tests
- Read the code. Complain. Rinse. Rather. Repeat.


tests/fixtures/tag.html
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
+<html>
+    <head>
+        <!-- Mobify Tag External -->
+        <script src="/tag/v7.js"></script>
```
jansepar, line 5: What scenario are you envisioning this being used for? This wouldn't work for the capturing case due to leakage.

tag/v7.js
```diff
@@ -0,0 +1,184 @@
...
+            'class': "mobify",
+            onerror: function() {self.disable()},
+            onload: postloadCallback
+        });
+    };
+
+    if (opts.capture) {
+        self.startCapture(load);
+    } else {
+        load();
```
jansepar, line 252: If we're not capturing, should we throw an async flag on the script? I wouldn't want us to synchronously block a page when dynamically loading the script.

jansepar:
Just head a brief look at this - it's looking good! Couple first few comments/thoughts:
- We talked about always loading a.js. Was this purposely not included, or is it just a TODO for you?
- We also talked about adding event listeners for DOMContentLoaded and onload in the tag itself, but that isn't implemented yet. Is that also in your TODOs? It would be great to get a list of TODOs that haven't been implemented yet.

tag/v7.js
```diff
@@ -0,0 +1,184 @@
...
+    self.options = options;
+
+    self.debug("Init Called");
+
+    if (self.isDisabled()) {
+        self.debug("Tag is disabled.");
+        return;
+    }
+
+    if (self.isPreview()) {
```
jansepar: It would actually be nice to be able to have "canPreview" in the options passed. That way we can nullify the use of preview if a user doesn't want it. If we did that, maybe we could even do away with this concept of open source vs non-open source tag. One tag to rule them all?

tag/v7.js
```diff
@@ -0,0 +1,184 @@
...
+
+// Mobify.init initializes the tag with the `options`.
+//
+// Format of `options` object:
+// Mobify.init({
+//     getMode: function(Mobify) {
+//         // Return mode based on device or other settings.
+//         // `mode` is a key in to the options object
+//         // that selects our device-specific options.
+//         return 'desktop'
```
jansepar, line 179: Currently, getMode is just a method that sets a variable that allows you to select a configuration object (which also must be declared in `init`). What are your thoughts around doing it this way vs just allowing users to override a "getOptions" where they could write the logic themselves. For example:

```
getOptions: function(Mobify) {

    if (/ipad|foo/.test(window.navigator.userAgent) {
        return {
             url: "http://cdn.mobify.com/foo/mobify.js",
             capture: true
        }
    }
}
```

Not sure if I have a preference yet, just wondering about your thinking here.

jansepar:
@rrjamie I'm feeling good about stage 1 of this tag. Shall we move on to stage 2 and 3?

tests/fixtures/tag.html
jansepar, line 5: (this is for testing purposes only)

tag/v7.js
```diff
@@ -0,0 +1,200 @@
...
+
+    // Reload the page (location.reload has problems in FF)
+    window.location = window.location.href;
+};
+
+// Mobify.collectTiming collects DOMContentLoaded time,
+// and Load time.
+Mobify.collectTiming = function() {
+    if (window.addEventListener) {
+        window.addEventListener('DOMContentLoaded', function() {
```
jansepar, line 106: Just had a thought - this DOMContentLoaded event isn't always the one we're going to want to use. When transforming a page, this will fire once the HTML has downloaded from the server, not when the page has document.opened, document.written, and scripts loaded. We want to record these numbers when _not_ capturing so that will have the ability to record DOMContentLoaded and onload when _not_ running mobifyjs.

tag/v7.js
```diff
@@ -0,0 +1,200 @@
...
+
+    var postloadCallback = function() {
+        if (opts.postload) {
+            Mobify.debug("Post Load Callback Firing");
+            opts.postload(self);
+        }
+    }
+
+    var load = function() {
+        preloadCallback();
```
jansepar, line 237: I'm trying to think of the need for this callback and can't really think of one... I'm not sure we should add functionality unless we can think of a legitimate use case. Can you think of any?

tag/v7.js
```diff
@@ -0,0 +1,200 @@
...
+        return;
+    }
+
+    var preloadCallback = function() {
+        if (opts.preload) {
+            opts.preload(self);
+        }
+    }
+
+    var postloadCallback = function() {
```
jansepar, line 229: Up until now, I've really been calling this "main executable". What do you think of "main" vs "postload"? Previously I called it "main" because this was the main entry point for where you were executing code (if you didn't combine the library and main files, as done in most library files).

examples/capturing-grumpycat/index.html
```diff
@@ -1,35 +1,64 @@
...
-        Mobify.Capture.init(function(capture){
-            // Grab reference to the captured document in progres
-            var capturedDoc = capture.capturedDoc;
-            var grumpyUrl = "/mobifyjs/examples/assets/images/grumpycat.jpg"
+        return "all";
+    },
+    all: {
+        capture: true,
+        url: "/mobifyjs/build/mobify.js",
+        postload: function() {
```
jansepar: I just went through the exercise of using this tag with the grumpycat example, and I have a few new thoughts:
- Our regex for determining blanket mobify.js support is quite long, and I think we would want it minified, but would we just minify the method within `init` itself? That seems wrong to me. Maybe the detect is just another method inside the tag itself?
- It seems strange for us to require typing `capture: true`. What if it was default `capture: true`, and you can still override it with `false` if you wish?
- The interface seemed strange to me - asking users to supply a mode that you must then match with a corresponding object literal as another argument to `init` does not seem like something that will be easy for users to grasp. What do you think about `getMode` being an optional parameter that we can still use, but isn't the default? I'm thinking something like this would be a better interface for most users:

```
Mobify.init({
    detect: Mobify.browserSupported,
    options: {
        url: "/mobifyjs/build/mobify.js",
        postload: function() {
            var capturing = window.Mobify && window.Mobify.capturing || false;
            if (capturing) {
                // Grab reference to a newly created document
                Mobify.Capture.init(function(capture){
                    // Grab reference to the captured document in progres
                    var capturedDoc = capture.capturedDoc;
                    var grumpyUrl = "/mobifyjs/examples/assets/images/grumpycat.jpg"

                    var imgs = capturedDoc.getElementsByTagName("img");
                    for(var i = 0; i < imgs.length; i++) {
                        var img = imgs[i];
                        var ogImage = img.getAttribute("x-src");
                        img.setAttribute("x-src", grumpyUrl);
                        img.setAttribute("old-src", ogImage);
                    }

                    // Render source DOM to document
                    capture.renderCapturedDoc();
                });
            }
            else {
                console.log('Executing post-capture!');
            }
        }

    }
});
```

Thoughts?

jansepar: Basically, I think the interface to `init` should be simple for basic use-cases, but advanced enough (using `getMode`, etc) for MPS use.

tag/v7.js
rrjamie, line 106: yeah, we aren't going to use them, but no reason to add a bunch of code to exclude them

tag/v7.js
rrjamie, line 252: I don't think that is necessary. It will load async because of the way we added it ot the page.

rrjamie:
re:  We talked about always loading a.js. Was this purposely not included, or is it just a TODO for you?

You can do this by using a `desktop` mode that loads a.js.

tag/v7.js
rrjamie, line 179: I figured it was easier to decouple them, because we can provided pre-baked `getMode` functions.

For the previous tags, since we had to do everything in a detector function, it was difficult for people to edit and understand. If we make it more declarative, common changes should be straight forward.

tag/v7.js
rrjamie, line 229: I like postload because it matches up with `preload`.

examples/capturing-grumpycat/index.html
rrjamie: Well, the most common use case is CST: we have a `mobile` mode, and a `desktop` mode, and sometimes `tablet` mode. So the goal here was to make it easy to handle all those different cases, because CST often has problems rolling out tags that deal with multiple devices.

Re: getMode:

Like my previous comment, I wanted to decouple the two because people often want to add new devices that load different bundles. So it'd be easy this way to provide

1) Pre-baked `getMode` for each different case.

2) It's easy to fill out the different keys declaritively.

Re: Default values:

1) I don't think it's work including a blanket regex in the tag, because most people will override it, and thus it's just wasted space.

2) We can make `capture: true`, the default, but it doesn't save that much space, and I prefer to be explicit. I could go either way on this one.

jansepar: I'm totally fine with `getMode`, and it works great for the CST use case - I just would like a simpler interface for use cases were we aren't needing to load stuff on tablet vs mobile vs desktop. If you look at the example in `grumpycat`, the call to `init` is much larger and more complicated then it needs to be. I'm thinking we could easily continue having `getMode`, as well as the example I defined above.

jansepar:
> You can do this by using a desktop mode that loads a.js.

Yep, that comment was quite old :)

jansepar:
> You don't have to do this in the tag?
> 
> The function will exist in Mobify.options

Ah cool, I'll fix up the code in Capturing.

tag/v7.js
jansepar, line 229: I could have sworn I wrote a comment about this - I was actually wondering where you think preload would be useful. I think you had an example but I can't remember what it was.

examples/capturing-grumpycat/index.html
rrjamie: Except that the only thing that is huge and ugly is `getMode`, so we'd move that as a default in to the tag as `browserSupported`, but then it's adding a bunch of weight to the tag.

We can try that, making a default `getMode` to the tag, that does what that does and returns says 'supported' and 'unsupported' as modes, then you can do (and I'd write it this way so it's not a mess, if we are doing the `main` thing):

```
<!-- Tag Goes Here -->
<script>
var myMain = function() {
    var capturing = window.Mobify && window.Mobify.capturing || false;
    if (capturing) {
        // Grab reference to a newly created document
        Mobify.Capture.init(function(capture){
            // Grab reference to the captured document in progres
            var capturedDoc = capture.capturedDoc;
            var grumpyUrl = "/mobifyjs/examples/assets/images/grumpycat.jpg"

            var imgs = capturedDoc.getElementsByTagName("img");
            for(var i = 0; i < imgs.length; i++) {
                var img = imgs[i];
                var ogImage = img.getAttribute("x-src");
                img.setAttribute("x-src", grumpyUrl);
                img.setAttribute("old-src", ogImage);
            }

            // Render source DOM to document
            capture.renderCapturedDoc();
        });
    }
    else {
        console.log('Executing post-capture!');
    }
};

Mobify.init({
    // No getMode because we are using the default
    supported: {
        url: "/mobifyjs/build/mobify.js",
        postload: myMain
    }
});
</script>
```

tag/v7.js
rrjamie, line 252: re: preload (commented on the wrong line) 

setting a background color, a spinner, etc.

tag/v7.js
rrjamie, line 229: setting a background color, a spinner, etc.

examples/capturing-grumpycat/index.html
jansepar: I like it! Although I'm not huge on the variable name `supported`. Reason is because "supported" is supposed to imply "if supported, then use these options", but I really think that is implied. I think instead, we already have an object passed to `init`, we could just define `url` and `postload` in that object directly. For example:

```
Mobify.init({
    // No getMode because we are using the default
    url: "/mobifyjs/build/mobify.js",
    postload: myMain
});
```

Thoughts? And did you want to make the modifications in the code with the default `getMode` stuff or should I?

rrjamie: I don't really like having those options top level because it does not extend easily to multiple modes.

tag/v7.js
jansepar, line 252: Right, that's a good potential use case!

examples/capturing-grumpycat/index.html
jansepar: Hmm, I prefer it because we would just have an object within an object, which would be kind of redundant. Since the `getMode` feature is really just a way to allow you to select different options, I figure that if you're not using `getMode` then your options should just be set on the top level.

jansepar:
How do you feel about adding everything to `Mobify.Tag` so we don't full up the `Mobify` object with stuff? It would be similar to how we did it with other libraries, and it would keep the `Mobify` object relatively clean.

rrjamie:
re: Mobify.tag

sure

jansepar:
Is the tag setup yet to have a way of being in "disabledButStillLoadMobifyForAnalytics" mode? How do you envision that will look? I'm guessing set a cookie and check for that in the `getMode` method?

rrjamie:
re: MPS

You would have "enabled", "disabled", and "unsupported". You'd have getMode() return "enabled" or "unsupported".

In mobify.js, if we want to disable, we set the mobify-mode cookie, and refresh.

tag/v7.js
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    // if the "options" objects has a mode, grab the mode and return the options
+    // set for that mode
+    if ('getMode' in options) {
+        var mode = self.getCookie("mobify-mode") || options.getMode(Mobify);
+        return options[mode];
+    // if there is no mode set, return the options object if the browser is
+    // supported, or if we're not capturing
+    } else if (options.capture === false || Mobify.supportedBrowser()){
```
rrjamie: I find this non-intuitive.

Why don't we just have a `default` mode keys then that we will use if `getMode` doesn't exist instead of returning the root object. Or, have an `enabled`, `disabled` default mode keys.

Also, why check options.capture == false? When is someone ever going to use Mobify.js if the only way they are using is not capturing?

src/capture.js
```diff
@@ -451,23 +451,25 @@ Capture.getMobifyLibrary = function(doc) {
...
+ * Grabs the postload function/src/script if it exists
  */
-Capture.getMain = function(doc) {
+Capture.getPostload = function(doc) {
     var doc = doc || document;
-    var mainScript = undefined;
-    if (window.Mobify && window.Mobify.mainExecutable) {
+    var postloadScript = undefined;
+    // mainExecutable is used for backwards compatibility purposes
+    var postload = window.Mobify && (window.Mobify.getOptions && window.Mobify.getOptions().postload.toString() || window.Mobify.mainExecutable);
```
rrjamie: `getOptions()` might not have a `postload` and this could blow up

tag/v7.js
jansepar: I just considered the root object the `default` keys. I see what you're saying about it being counter intuative in that we should always be calling `getMode`, even if no `getMode` is defined. I just want to avoid passing a named object into the options object.

I would prefer this (when not overriding `getMode`):

```
Mobify.init({
    url: '/foo.js'
    postload: function(){ return 'foo' }
})
```

Over this, which in my opinion would be redundant:

```
Mobify.init({
    default: {
        url: '/foo.js'
        postload: function(){ return 'foo' }
    }
})
```

But I'm okay with re-gigging this function to always call `getMode`.

src/capture.js
jansepar: Good catch!

tag/v7.js
rrjamie: I know it kinda looks redundant, but I prefer that over making it confusing which object gets used.

rrjamie: Also, I don't think we always have to include a `getMode` if we assume that the mode is `default` if `supportedBrowser` returns true.

jansepar: Hmmm... That's a tough one. I guess you're wanting to avoid confusion from a code perspective? I don't think the way I suggested would have confusion from a user perspective.

Also, for your other question of "When is someone ever going to use Mobify.js if the only way they are using is not capturing?", you can use Mobify.js today to have responsive images without capturing. Plus in the future MPS might have many features and someone might use a combination of features that doesn't require capturing. Plus we also plan to load mobify.js in non-capturing mode when wanting to run Grog in the "disabled" case.

The check for `options.capture == false` is there to make the default for `capture` to be `true`. That could probably be done better by setting `capture = options.capture || true` up at the top instead.

rrjamie: Re: capture: false.

Well, if you are doing it with the root object, that implies a single mode, so then you are doing

```
Mobify.init({
    capture: false,
    url: "foo.js"
})
```

but why would you just a mobify tag at this point? You can just load `foo.js` directly. So, that clause has no purpose.

jansepar: The way I defaulted `capture` to `false` isn't just for single mode - it also defaults `capture` to `true` when using `getMode` as well.

jansepar:
Alright, this is looking good! :+1:

PR #239, opened 2014-01-07
Make Tag Backward Compatible

rrjamie:
**Status**: _Opened for Visibility_

**Reviewers**: @jansepar 
**Linked PRs**: https://github.com/mobify/portal_app/pull/2203 (This has to be deployed before we roll out tags)
## Changes
- Adds tests for v1.0 Projects
  - Fixes issue on v1.0 projects (Mobify.tagVersion must exist)
- Adds test for v1.1 Projects
- Adds test for v2.0 Projects
- Adds test that loads tag in old browser, and ensures it does not blow up.
  - Fixes issue with `console.log` and `Date.new` not existing on IE 
- Adds support for preview
## How to test-drive this PR
- Run the tests, and then run them on sauce labs.
- Use Charles to inject the tag in a page (eg, from a CST project) and map preview.mobify.com to your local portal_app
  - Ensure that old projects load (1.0, 1.1, MPS)
  - Ensure you can preview


tests/tag.html
```diff
@@ -27,6 +27,32 @@
...
+            return false;
+        }
+        // match[5] == IE
+        if (match[5] && +match[6] < 6) {
+            return false;
+        }
+
+        return true;
+    }
+
```
jansepar: We do we need this? This method is already part of the tag itself is it not?

jansepar: If this is specifically for the 1.1 and 1.0 tests, then we should change the regex match here because those two versions support a much smaller subset of browsers.

tests/fixtures/tag/project2-0.html
```diff
@@ -0,0 +1,25 @@
...
+    <head>
+        <script src="/tests/utils/proxyassert.js"></script>
+        <!-- Mobify Tag External -->
+        <script src="/tag/v7.js"></script>
+        <script>
+        Mobify.Tag.init({
+            getMode: function() { return 'mobile' },
+            mobile: {
+                capture: true,
+                url: "/tests/fixtures/tag/project2-0/mobify.js"
```
jansepar: I'm not sure if this is something we want to do - it's fine for 1.1 and 1.0 because those aren't changing much, but 2.0 is constantly changing, and I think it's valuable to always run the latest version of 2.0 against this test. Instead of committing a built file, how about just changing this to `/mobifyjs/build/mobify.js`, and the stuff that you have at the bottom of the build file that calls `capture.init` can go into `postload` in this tag.

tag/v7.js
```diff
@@ -57,24 +60,27 @@ Tag.startCapture = function(callback) {
 };
 
 // Tag.getCookie fetches the values of a cookie by the given `name`
+// Returns `undefined` if no cookie matches.
 Tag.getCookie = function(name) {
-    var re = new RegExp("(^|; )" + name + "=([^;]*)");
+    // Internet Explorer treats empty cookies differently, it does
+    // not include the '=', so our regex has to be extra complicated.
+    var re = new RegExp("(^|; )" + name + "((=([^;]*))|()(;|$))");
```
jansepar: Instead of `|()`, couldn't we just do `?`? So like `"((=([^;]*))?(;|$))"`. Then lower down, couldn't you do `(typeof match[4] === 'string' ? match[4] : '');`?

rrjamie: mine is shorter :)

tests/tag-old-browser.html
```diff
@@ -0,0 +1,50 @@
...
+            start();
+        };
+
+        Mobify.Tag.init({
+            getMode: function() {return 'desktop'},
+            desktop: {
+                capture: false,
+                url: "/tests/fixtures/tag/desktop.js"
+            }
+        });
```
jansepar, line 45: It would also be nice to have a test the has `capture: true` and see if we get the same outcome (nothing breaking).

rrjamie, line 45: I believe this is covered by the `single-mode` test/fixture.

jansepar, line 45: But isn't this file the only one that runs on old browsers?

tests/tag.html
rrjamie: done.

tests/fixtures/tag/project2-0.html
rrjamie: done.

tag/v7.js
rrjamie: done.

tests/tag-old-browser.html
rrjamie, line 45: done

rrjamie:
intentional. If there's a `;`, then it's not the last cookie, otherwise it's just at the end (`$`).

jansepar:
@rrjamie could we get a test around `getCookie`?

rrjamie:
We could test that directly, but I believe it's adequately tested by the integration tests that hit `isDisabled`.

jansepar:
Should we remove the `/v7/` from `Tag.previewUrl = "https://preview.mobify.com/v7/";`? Doesn't look like we need a special path anymore. It still works the way it is because of the Preview app handling `/*` the same way everywhere, but still, we should remove it to avoid confusion.

rrjamie:
I want to keep it, just in case we have to special case it in the future.

jansepar:
^^^ fair enough! Thats a good idea.

:+1:

PR #240, opened 2014-01-14
Make sure we pass prefix to enable method

MikeKlemarewski:
Fixes a call to enable inside Capture.render

**Status**: _Ready to Merge_

**Reviewers**: @jansepar 
## Changes
- pass this.prefix to enable so we properly re-enable resources
## How to test-drive this PR
- run test suite `grunt test`
- run Capture.render('')  which should change the x-src to src


jansepar:
:+1:

PR #241, opened 2014-01-15
Fix options in docs

jansepar:


noahadams:
seems reasonable that the docs match the implementation... :+1:

PR #242, opened 2014-01-20
Make Tag Small

rrjamie:
**Status**: _Review_
**Reviewers**: @jansepar 
**JIRA**: [PROD-1266](https://mobify.atlassian.net/browse/PROD-1266)
## Changes
- Adds JSDoc comments to aid in minification
- Rewritten to be slightly more amendable to minification
- Renamed some variables to `pre`, `post`, and `mode` to save a little bit of space.
- Split up tests to run on the full minified, and the mostly-minified (exposed) forms.
## How to test-drive this PR
- Runs tests (locally, and on saucelabs)
- Create tag and test it
## Research resources
- https://developers.google.com/closure/compiler/docs/js-for-compiler


tag/v7.js
```diff
@@ -241,22 +369,22 @@ Tag.init = function(options) {
         preloadCallback();
 
         var options = {
-            id: "mobify-js",
-            src: opts.url,
-            'class': "mobify",
-            onerror: function() {self.disable()},
+            id: "mobifyify-js",
```
jansepar: mobify-ify?

jansepar:
@rrjamie just going to continue our conversation here. Here is your comment:

> we can save another 200-300 bytes if we don't include supportedBrowser in the tag we ship. I see two options (1) Use one tag in MDS/MPS/Cloud and put a different one on the website (2) Remove that code, and instruct users to add a if (/foo/.test(ua)) { Mobify.Tag.init(…) } around their initialization code

For option #2 - couldn't we do that even for CST projects? Instead of a `getMode` method, we could just people create a chain of `if else` conditions that initalized the tag in different ways? I guess the advantage of `getMode` way is that the declarative way is a bit more concise. Basically its this:

_Option A_

```
if (desktop) {
    Mobify.Tag.init({foo: 'bar'})
} else if (mobile) {
    Mobify.Tag.init({bar: 'baz'})
}
```

**vs**

_Option B_

```
Mobify.Tag.init({
    getMode: function(){
        if (desktop){
            return 'desktop';
        }
        else if(mobile) {
            return 'mobile';
        }
    }
    mobile: {foo: 'bar'}
    desktop: {bar: 'baz'}
});
```

So, right now what you're advocating is that for the basic mode, you'd go with _Option A_, and for CST projects, you'd go with _Option B_. I personally feel like it would make sense to stick to be consistent and just stick with one of the options for both use cases. Agreed or disagreed?

And assuming you prefer _Option B_, why do you think it's a better approach then _Option A_?

rrjamie:
Disagree, because we need the mode switching logic to support Enabled/Disabled modes in MPS, and in theory, Mobile/Desktop modes for CST. All modes have to be declared in the init call. Otherwise, they have to implement that logic as part of their tag options themselves.

However, for people that just want to use a single mode, they can go

if (mobile) {
   Mobify.Tag.init({url: "foo.js"})
}

jansepar:
So, as per our discussion, what if we did this:

```
(function(window, document, Mobify){Tag = {}....blah..blah..}(window, document, Mobify)); <-- all tag logic
var supportedBrowser = /iphone|android.*mobile/.test(navigator.userAgent);
if (supportedBrowser) {
    Mobify.Tag.init({url: "foo.js"})
}
```

And we would generate two different versions of the tag - one with `supportedBrowser` compiled with it, and one without. That way, if someone DOES want to define their own "supportedBrowser" method, they can easily delete the old one so that they can limit the size of the tag.

rrjamie:
Sure, we can do that. But I think it's better that it just live with the documentation then as one way to use the tag.

jansepar:
@rrjamie hmmm actually if we do it that way then it won't exactly be easy to test the `supportedBrowser` method. We could just have two different tags, and in the docs just say "Use this tag if you want to target all browsers", and "Use this tag if you want to override browser detection or do more complicated things"

rrjamie:
Or we just move that in to a separate test file and copy-paste-it when it changes.

jansepar:
> Or we just move that in to a separate test file and copy-paste-it when it changes.

:( I'd like to avoid that if possible...

jansepar:
@rrjamie any further thoughts on how we want to manage where the detector function goes? It would be great to get this merged.

jansepar:
Another thing we could do is just have a separate "supportedBrowser.js" that has it's own set of unit tests. Then for tags that don't define a "mode", you just include that script yourself and do it the way I mentioned last time:

```
(function(window, document, Mobify){Tag = {}....blah..blah..}(window, document, Mobify)); <-- all tag logic
var supportedBrowser = /iphone|android.*mobile/.test(navigator.userAgent);
if (supportedBrowser) {
    Mobify.Tag.init({url: "foo.js"})
}
```

rrjamie:
I've updated the code to move supportedBrowser.js out.

@jansepar do you want to update the docs/examples?

tag/v7.js
```diff
@@ -1,67 +1,168 @@
...
+
+    To Compile For Production:
+        java -jar <path to compiler.jar> \
+        --accept_const_keyword \
+        --compilation_level ADVANCED_OPTIMIZATIONS \
+        --define EXPOSE=false \
+        --define SINGLE_MODE=false \
+        --define ALLOW_SKIP_PREVIEW=false \
+        --js tag/v7.js \
+        --use_types_for_optimization > tag/v7.min.js
```
jansepar, line 30: Since we've moved the browser support stuff into a separate file, I actually don't think we need to create a separate tag for open source and production. The differences here (`SINGLE_MODE` and `ALLOW_SKIP_PREVIEW)` aren't really necessary to override, and in my option not worth having to maintain two tags for.

For `SINGLE_MODE`, I don't think we need this variable at all - checking for the existence of `options['mode']` is good enough. Same goes for `ALLOW_SKIP_PREVIEW`. There isn't anything in the way of saving space - we could just remove both these variables and save even more bytes!

rrjamie, line 30: Uhh, it does save some bytes, as they all get compiled out.

Removing SINGLE_MODE from production tag saves about 20 bytes. Removing skipPreview saves about 12 bytes.
Since we'll never use those features in MPS/MDS, we can remove them; I don't think it's too huge of a burden 'cause we going to do it once and then forget about it for another year.

rrjamie, line 30: though, I could go either way on this

jansepar, line 30: Ah, yeah you're right - I assumed the override was simply setting some variables, but the compiler does go through and make things smaller due to the fact that setting those vars to `false` means it can strip out more code.

I would personally rather just have one tag for simplicity sake - the difference in size is quite minor, and I'd rather there be one tag to rule them all, rather then having a "open source" tag vs a "cloud" tag. Removing the need to have two separate tags was a mini-goal of mine for this new tag work.

jansepar:
I'm thinking of moving the new tag into the examples and docs once we're live on cloud - by then we should have ironed out any kinks (instead of having to update everything again if we want to make any changes, although the changes of that seem small).

jansepar:
:+1:

PR #243, opened 2014-02-03
First attempt at converting mobifyjs from using requirejs to using browserify! Looking good so far :)

jansepar:
This PR is mostly for funzies, so we can have a debate about whether requirejs or browserify makes the most sense for this project!


jansepar:
Closing this in favour of PR for switching to UMD rather then only using CommonJS definitions (https://github.com/mobify/mobifyjs/pull/248)

PR #248, opened 2014-02-22
AMD/Require.js to UMD/Browserify

jansepar:
**Status**: _For Review_
**Reviewers**: @noahadams
## Changes
- Replaces AMD definitions in each module with UMD definitions to support both Require.js and Browserify.
- Uses Browserify rather then Require.js to remove dependance on the Almond.js loaded.
- Reduces size of minified non-gzipped build from 24.6KB to 21.6KB
## How to test-drive this PR
- Ensure that all tests work, and try on a real site
## Notes:
- Motivations for this change:
  - Move away from needing to include almond.js to reduce size of build. As well, including almond.js makes no sense for mobify.js as nothing is loaded asynchronously. Browserify fits this bill better.
  - Why we didn't go with amdclean: it's use of esprima does not give a 1 to 1 mapping of generated file to individual built file.
  - Uses UMD rather then moving to CommonJS-only for a few reasons:
    - Keeping AMD definition ensures tests still work since the modules can be loaded via require.js
    - Keeping AMD definition ensures that any libraries that requires in modules separately still work, preserving backwards compatibility.


Gruntfile.js
```diff
@@ -52,28 +52,14 @@ module.exports = function(grunt) {
                 }
             }
         },
-        requirejs: {
```
noahadams, line 56: this PR is pretty opinionated, I'm tempted to bring it downa  notch by leaving the require build rules in as defaults but also allowing browserify, thoughts?

jansepar, line 56: I'm cool with leaving the requirejs stuff in the build file, but only one of them can be the default for `grunt build`. I'd want the default to be browserify.

noahadams:
:+1: your changes, do you :+1: mine?

jansepar:
:+1:

PR #250, opened 2014-02-26
PROD-1562: Detect whether to resize images from the viewport meta tag

jvaill:
The way we calculate the dimensions for resizing images does not work for non-responsive websites. Disable resizing for websites that are not responsive (that don't have an ideal viewport meta tag).

Status: **Opened for review**

Reviewers: @jansepar 
JIRA: **https://mobify.atlassian.net/browse/PROD-1562**
## Changes
- Adds a `resize` option to Image Resizer that serves images through the IR backend without resizing them
- Auto detects whether to resize images based on the document's viewport meta tag
## How to test-drive this PR
- Run the tests and examples


build/mobify.js
```diff
@@ -1568,32 +1578,39 @@ ResizeImages.processOptions = function(options) {
...
     }
+    
+    if (!opts.format && opts.webp) {
+        opts.format = "webp";
+    }
 
-    var dpr = opts.devicePixelRatio || window.devicePixelRatio;
+    // With `passthrough` images are served through IR without resizing,
+    // when set ensure that any options pertaining to resizing are cleared
+    if (opts.passthrough) {
```
jansepar: I'm not sure about "passthrough". If I was a user and I saw this parameter, I would assume "passthrough" means "pass through IR but literally do nothing", meaning I would assume it would also not change the format (to WEBP for example). "Passthrough" to me doesn't imply widths will be passed through.

Some alternatives: "passthroughWidth", or `"width": null`. I think I would prefer the latter

jvaill: I think `width: null` is confusing because the default behaviour of `width: undefined` resizes the image according to the browser's viewport, I'd expect similar behaviour with null.

Plus, I think we have to be careful about naming this something that associates it strictly with `width` because logically this disables resizing all-together (incl. other properties that control resizing today and in the future: `maxHeight`, etc.)

How about something like `resize` or `resizePassthrough`?

jansepar: I like `resize`. Default set to true, setting it to false will cause it to not use width detection. +1 to that

build/mobify.js
```diff
@@ -17,7 +16,8 @@ var requirejs, require, define;
         config = {},
         defining = {},
```
jansepar, line 17: Hmmm which version of grunt-contrib-requirejs do you have? Seems like you have a newer one since the almond version here is different

jvaill, line 17: I'm at 0.4.0 which is the minimum package.json supports, the almond dependency is at 0.2.9 though - should I downgrade?

jansepar, line 17: I don't mind upgrading as long as we lock down the version so that in the future we don't keep going back and forth between almond versions!

build/mobify.js
```diff
@@ -1559,41 +1607,95 @@ ResizeImages._getBinnedDimension = function(dim) {
...
-
-    var screenSize = Utils.getPhysicalScreenSize(dpr);
-
-    // If maxHeight/maxWidth are not specified, use screen dimensions
-    // in device pixels
-    var width = opts.maxWidth || ResizeImages._getBinnedDimension(screenSize.width);
-    var height = opts.maxHeight || undefined;
+    // A null value for `resize` (as is the default) triggers the auto detect
+    // functionality. This uses the document to determine whether images should
+    // be resized and sets a new default.
```
jansepar: Might want to mention that if there is no document present (which is allowed), we will continue with automatically resizing.

noahadams: That might not be appropriate behaviour for integration with 1.1, if for example, the reszie is being done on an element created with jquery and never bound to a document.

jansepar: @noahadams that should be fixed when @jvaill implements what we talked about earlier (if there is no document, assume we're going to resize).

jansepar:
:+1:

PR #257, opened 2014-03-18
Uses `window.location.reload()` in the v7 tag

jansepar:
Use reload on error

Reviewers: @rrjamie 
## Changes
- Use `window.location.reload()` instead of `window.location = window.location.href` in order to fix bug for sites that have hash params.
## How to test-drive this PR
- `grunt serve` and ensure tests pass: http://localhost:3000/tests/tag.html
- Put tag in a page, and specify a url to a path that doesn't resolve when loading mobify.js, and load that tag with a hash in the URL.


rrjamie:
+1

PR #259, opened 2014-03-20
Fixing an issue with grabbing head contents when tag is above <head>

jansepar:
One of my crazy 2:30am pull requests.


johnboxall:
:+1:

PR #260, opened 2014-03-24
PROD-1762: Adds `mobify-capture=false` as a parameter for preview.

rrjamie:
Allows users to disable capturing while loading preview.

Status: **Opened for Review**

Reviewers: @jansepar 
JIRA: https://mobify.atlassian.net/browse/PROD-1762
Linked PRs: **PORTAL PR TO BE DONE**
## Changes
- Adds `mobify-capture=false` flag.
## How to test-drive this PR
- Run the tests


rrjamie:
I'm not sure if I like this change. How many sites are going to being not-using capturing, but we still want them to be end-user testable?

We're smart enough to be able to test with Charles, so I don't see a strong use case for this change.

jansepar:
Agreed. Let's not do this for now.

PR #265, opened 2014-04-11
Restore capturedDoc

MikeKlemarewski:
Update the restore function to render the capturedDoc

**Status**: _Ready to Merge_

**Reviewers**: @jansepar 
## Changes
- Removes doc.write in favour of render
- Uses our waitForReady method instead of domIsReady
## How to test-drive this PR
- Check out a project and point the version of mobifyjs to this commit (easily done with the adaptivejs generator)
- make the project throw an error that results in a capture.restore() call
- make sure the original document is rendered


jansepar:
:+1:

PR #266, opened 2014-04-14
Render self.all() instead of captured doc + test

MikeKlemarewski:
Use self.all() in restore because we might have modified the capturedDoc

**Status**: _Ready to Merge_

**Reviewers**: @jansepar @noahadams 
## Changes
- Passes self.all() to render so we don't render the captured document
- Add a test
## How to test-drive this PR
- Checkout the repo
- `grunt test`


build/mobify.js
```diff
@@ -1653,8 +1653,8 @@ Capture.setElementContentFromString = function(el, htmlString) {
      * Therefore, bodyContent will have these tags, and they do not need to be added to .all()
      */
     captured.all = function(inject) {
-        return this.doctype + this.htmlOpenTag + this.headOpenTag + (inject || '') + this.headContent + this.bodyContent;
-    }
+        return this.doctype + this.htmlOpenTag + this.headOpenTag + (inject || '') + this.headContent + this.bodyOpenTag + this.bodyContent;
```
noahadams, line 1656: might be cool to offer the opportunity to inject markup into body as well

jansepar, line 1656: We went down that route, but then we realized we didn't have the option of injecting something at the end of body, because `bodyContent` includes `</body></html>`.

jansepar:
Btw another example of where tests are awesome - uncovered a bug in `all` :)

build/mobify.js
jansepar, line 1656: We could allow people to inject at the top of body, but that's not very useful.

noahadams, line 1656: might be useful to throw in a div or something, but i suppose you could do that with a script injected in the head

noahadams:
looks legit... :+1:

build/mobify.js
jansepar, line 1656: Yeah exactly. If there is ever a scenario when we need it, we will add it then. Any other issues with this PR or can we get it merged?

PR #269, opened 2014-04-15
v1.3: Don't lookup non-existent locale's in the i18n module

tedtate:

Reviewers: *\* @jansepar @rrjamie @noahadams **
## Changes
- Guard against looking up non-existent locales in the i18n function
## How to test-drive this PR
- Build yourself a mobify-client using this commit for mobify 1.3
- Create a project where you initialize an i18n module
- Add a locale to the i18n module which doesn't exist in the json source file
- Try looking up a key, it should return the value of that key under "DEFAULT" instead of erroring


jansepar:
:+1:

PR #270, opened 2014-04-15
V1.1 project name v7 tag fix

jansepar:

Reviewers: @tedtate @rrjamie 
JIRA: None
## Changes
- Makes sure Mobify.ajs is defined, even if the tag doesn't have it.
## How to test-drive this PR
- For a site that has a v7 tag, load a bundle and make sure Mobify.ajs is defined.


tedtate:
+1

joaquin-up

PR #273, opened 2014-05-09
Do not patch anchor links on Firefox 29 or above

jansepar:
Status: For review

Reviewers: @noahadams 
## Changes
- The anchor patch is no longer necessary on newer firefox browsers. This turns the patching mechanism off for 29 and above.
## How to test-drive this PR
- Go to http://localhost:3000/tests/fixtures/anchor-test.html and run test 1. On older firefox browsers, when you click test 1, there should be _no_ hash in the URL. In newer Firefox browsers, there _should_ be a hash in the URL. Same with non-Firefox browsers.


src/patchAnchorLinks.js
```diff
@@ -16,10 +16,18 @@
 }(this, function (Utils) {
     var exports = {};
 
-    var isFirefox = function(ua) {
+    var isOldFirefox = function(ua) {
```
noahadams, line 19: this bad boy is just crying out for a test

noahadams:
@jansepar have a look at my changes, I had to fix up the regex and rejigger the isOldFirefox functiona  bit to make it testable.

jansepar:
:+1: to your changes!

jansepar:
Do you :+1: mine?

noahadams:
:+1:

PR #275, opened 2014-05-15
Prevent `captured` from leaking onto global

jansepar:
If only we had linting on - this stuff would have never happened :). Next on the TODO list.


dbader:
@jansepar Can we run this with 'use strict'? That also would've caught it I think.

jansepar:
@dbader good call! I won't do it for this PR, because if I was to add 'use strict' I'd want to test it more thoroughly. When I add linting I will add 'use strict' :)

noahadams:
:+1:

PR #276, opened 2014-05-16
Prefer the full open <body> when unmobifying.

johnboxall:
@jansepar you the man now :dog:


unmobify.js
```diff
@@ -311,7 +311,7 @@ var tagVersion = parseInt(Mobify.tagVersion && Mobify.tagVersion[0] || Mobify.co
             // If used as part of the API, content will be ready, otherwise find it.
           , headMarkup = Mobify.headMarkup
           , bodyMarkup = Mobify.bodyMarkup
-          , bodyOpenTag = openTag(bodyEl)
+          , bodyOpenTag = Mobify.bodyOpenTagMarkup || openTag(bodyEl)
```
jansepar, line 314: So yeah, the code inside the `if` that follows this doesn't execute and because of that, `bodyOpenTag` doesn't get properly set.

jansepar:
:+1: looks good!

PR #285, opened 2014-09-25
Adding comment clarifying the use of "reload"

jansepar:
I was super confused as to why we used reload here, as I thought it would break FF, despite the fact that I wrote this code. Realized that it wouldn't though - adding a comment for clarification to others and my future self! Just another testament to the importance of good commenting :)


jansepar:
Great job Shawn :+1:

PR #289, opened 2014-09-27
V1.1 - GOLD-883: iOS Scroll Fix V2

jansepar:
Merge checklist:
- [x] Removes the old iOS 8.0 patch
- [x] Adds the new iOS 8.0 patch
- [x] Uses the correct iOS 8.0 regex (that works with the simulator)
- [x] Calls the new patch
- [x] Has a test for the new patch
- [x] Has a test for the iOS 8.0 form sibblings bug

### DIFF
--- a/api/main.js
+++ b/api/main.js
@@ -107,13 +107,8 @@ $.extend(Mobify.transform, {
         };
 
         if (Mobify.isIOS8_0()){
-            // See `utils.js` for more information about this fix.
-            Mobify.ios8_0ScrollFix(document);
-            window.requestAnimationFrame(function(){
-                window.requestAnimationFrame(function(){
-                    write();
-                })
-            })
+            // See `util.js` for more information about this fix.
+            Mobify.ios8_0ScrollFix(document, write);
         } else {
             write();
         }
--- a/api/main.js
+++ b/api/main.js
@@ -92,6 +92,11 @@ $.extend(Mobify.transform, {
             timing.logPoints();
         }
 
+        if (Mobify.isIOS8_0()){
+            // See `utils.js` for more information about this fix.
+            Mobify.ios8_0ScrollFix(document);
+        }
+
         // We'll write markup a tick later, as Firefox logging is async
         // and gets interrupted if followed by synchronous document.open
         window.setTimeout(function(){
--- a/api/main.js
+++ b/api/main.js
@@ -92,22 +92,31 @@ $.extend(Mobify.transform, {
             timing.logPoints();
         }
 
+        var write = function(){
+            // We'll write markup a tick later, as Firefox logging is async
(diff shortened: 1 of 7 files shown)


api/main.js
```diff
@@ -92,17 +92,31 @@ $.extend(Mobify.transform, {
...
+
+                // In Webkit, `document.write` immediately executes inline scripts 
+                // not preceded by an external resource.
+                document.write(markup);
+                document.close();
+            });
+        };
+
+        if (Mobify.isIOS8_0()){
+            // See `utils.js` for more information about this fix.
```
scalvert: I believe the file is `util.js`

api/util.js
```diff
@@ -40,54 +40,22 @@ Mobify.isIOS8_0 = function() {
...
-
-    return htmlString.replace(BODY_REGEX, openingBodyTag + script);
+Mobify.ios8_0ScrollFix = function(doc) {
+    var meta = document.createElement('meta');
+    meta.setAttribute('name', 'viewport');
+    meta.setAttribute('content', 'width=device-width');
+    // Using `getElementsByTagName` here because grabbing head using
+    // document.head will throw exceptions in some older browsers
+    // (iOS 4.3).
+    doc.getElementsByTagName('head')[0].appendChild(meta);
```
johnboxall: @jansepar Probably pretty rare, but you might want to guard against the case when there is no `<head>` element.

jansepar: Pretty sure that is impossible. Even if you didn't have a head in your markup, the browser will create it for you.

But yeah, doesn't hurt to guard against it just incase :)

api/main.js
```diff
@@ -92,17 +92,31 @@ $.extend(Mobify.transform, {
...
+            Mobify.ios8_0ScrollFix(document);
+            window.requestAnimationFrame(function(){
+                window.requestAnimationFrame(function(){
```
jansepar: FYI, we'll probably refactor these calls to `requestAnimationFrame` into the `ios8_0ScrollFix` itself. Incase anyone was going to comment on that :)

tests/runner.sh
```diff
@@ -1,6 +1,12 @@
...
 PID=$!
+
 sleep 1
 phantomjs tests/phantom.js | grep '<*>' | tee report.xml
-kill $PID
\ No newline at end of file
+ret_code=$?
+
+kill $PID
+exit $ret_code
```
jansepar, line 10: What's the dealio with changing this?

jansepar:
:+1: looks good. Let's merged this once we get approval from QA. We're good to spread this out to 1.0, 1.3, and 2.0!

MikeKlemarewski:
Also tested this scroll patch on Android, iOS 7 and iOS 8. Seems to work!

PR #292, opened 2014-09-29
V1.3 - GOLD-883: iOS Scroll Fix V2

jvaill:
Merge checklist:
- [x] Removes the old iOS 8.0 patch
- [x] Adds the new iOS 8.0 patch
- [x] Uses the correct iOS 8.0 regex (that works with the simulator)
- [x] Calls the new patch
- [x] Has a test for the new patch
- [x] Has a test for the iOS 8.0 form sibblings bug


jansepar:
:+1:

PR #305, opened 2015-06-30
Desktop fall-thru issue

wizardlyhel:
## Status: Opened for visibility

We have been trying to solve this issue for awhile. Our clients are reporting that the mobile site have been randomly falling to the desktop state. We had no luck of reproducing this bug so we had no way to figure out what is it that the clients are seeing. It is only until now that we were able to reproduce this bug. Even if we can reproduce it, the chance of reproducing this bug is very low.
### Steps to reproduce the issue
1. VPN to a London access point
2. Navigate normally within mobified pages
3. If you are lucky, you will fall thru to desktop (Chance of 1 in 1000 pageviews)

**Note:** The above steps works with a preview bundle as well but will not work if it is local.
### Investigations

Due to the difficulty of how often we can reproduce this bug, I decided to place debug logging points throughout the code path. In summary, the page that is suppose to be mobified is falling thru because it did not found a matching template.

Further investigation revealed that in the `evalCollection` function, the `$.each` is not iterating the template object.

screen shot 2015-06-29 at 10 22 58 am

In the above screenshot, I was logging information from the `evalCollection` function.

`source` object is the template object that got pass in from the knof.

The first object with the empty template name is the correct state. Notice that it has 2 object under the source object and we have a boolean indicator `inEachFunction`, to indicate if we arrived in the callback function. `eachFragment` shows how many times that `$.each` callback function is iterated.

The second object with the redirect template name is the incorrect state. Notice that it has 2 objects under the source object, however, the `$.each` callback was never called which is indicated by the false value of `inEachFunction`.

I also make sure that `$.each` and any required objects exists before the `$.each` function.

This PR contains one of the variations of code that I was using to track down the root of the bug. This particular code variation does not fall thru to desktop, at least it hasn't yet.
### Observations

I have been running some number tests to create a baseline that we can compare.

**Without Fix:** Desktop fall thru happens within the range of 0 to 1000 pageviews

**With Fix:** Haven't seen fall thru yet. Highest pageview count is 8715.
### Next Steps
- [x] Test this code on production sites to validate that it does "fix" the desktop fall thru issue (This code is run again Christie's production for 2 weeks so far. We are seeing about maximum of 3 drop thrus comparing to 10-20 drop thrus before fix)

**Participants:** @johnboxall @jansepar @benlast @donnielrt @cbildfell 
## Changes
- Instead of using Zepto's `$.each`, we pull in the exact logic of Zepto's `$.each` to iterate any objects that doesn't have `.length` property

### DIFF
--- a/api/cont.js
+++ b/api/cont.js
@@ -116,8 +116,10 @@
                     .evalReference();
             };
 
-            if (typeof source.length === 'number') {
-                $.each(source, eachCallback);
+            if (typeof sourceLength === 'number') {
+                for (i = 0; i < sourceLength; i++) {
+                    eachCallback(i, source[key]);
+                }
             }
             else {
                 for (key in source) {
--- a/api/cont.js
+++ b/api/cont.js
@@ -103,8 +103,8 @@
             var source = this.source(),
                 sourceLength = source.length,
                 continuation = this;
- 
-            $.each(source, function(idx, sourceFragment) {
+
+            var eachCallback = function(idx, sourceFragment) {
                 if (sourceFragment && (sourceFragment.jquery || sourceFragment.nodeType)
                     && (typeof idx == "string") && idx.indexOf('$')) {
                     var root = continuation.root;
@@ -114,7 +114,17 @@
                 continuation
                     .extend({source: sourceFragment}, idx, sourceLength, value)
                     .evalReference();
-            });
+            };
+
+            if (typeof source.length === 'number') {
+                $.each(source, eachCallback);
+            }
+            else {
+                for (key in source) {
+                    eachCallback(key, source[key]);
+                }
+            }
+ 
             return value;
         }
 


api/cont.js
```diff
@@ -114,7 +114,17 @@
                 continuation
                     .extend({source: sourceFragment}, idx, sourceLength, value)
                     .evalReference();
-            });
+            };
+
+            if (typeof source.length === 'number') {
+                $.each(source, eachCallback);
+            }
+            else {
```
jansepar, line 124: @Helen-Mobify could you explain a bit about why this fixes the issue? Why is it that it's better to use the `for..in` when `source.length` is a number? Is Zepto's each function just completely busted on some devices?

wizardlyhel, line 124: I am not sure why Zepto's `$.each` fails to iterate but we did proved that it failed to iterate. All I did here was pulling out Zepto's `$.each` logic so that we are not using what is assign to `$.each`. This is just one of the code variation that I tried for in attempt to target the fault. We'll do proper cleanup once I verified this does fix the issue.

Referring to the screenshot I provided, you can see in the logging that `$.each` function worked in the previous template object iteration but failed on the second template object iteration. "When" `$.each` fails to iterate is random. This means that it is not failing at specific template object nor at specific position.

So we know for certain the following:
- `$.each` is assigned properly before the iteration
- We know the `$.each` callback never got executed
- We know that pulling out the exact logic of `$.each` doesn't fall thru to desktop (Waiting for a live test to verify this)

So how did the assignment of `$.each` randomly disappear? I really have no idea how I can dig even deeper than this.

wizardlyhel, line 124: It's not just on some device. It's across all devices and browser. We can see this in Google Analytics.

jansepar, line 124: So how did you think of even trying to stop using `$.each`? And also how did you know to only use it if `source.length` is undefined?

wizardlyhel, line 124: The reason why I tried not using `$.each` is due to the following reasons:
1. `$.each` is working on the previous template object
2. `$.each` is defined before the iteration
3. we never reach the `$.each` callback function

It would've make sense had `$.each` was failing to iterate every time, however, it was failing only sometimes. @benlast and I were thinking there may be some race condition here that made `$.each` failed sometimes. So we took out `$.each` just to see if we can reproduce the same error.

The `source.length` logic is grabbed from zepto's definition of `likeArray` function

Zepto's `$.each`: https://github.com/mobify/zepto/blob/master/src/zepto.js#L232
Zepto's `likeArray`: https://github.com/mobify/zepto/blob/master/src/zepto.js#L55

wizardlyhel:
Updates from the live test.

This code variation did not eliminate desktop fall-thru, however, it reduces the possibility down to 1 to 3 fall-thrus so far.

Base on client data, we can see that fall-thrus are happening at about 1% of the time. With this code variation, we reduce the fall-thru rate to 0.1% of the time. We don't have enough data to confirm that the number of fall-thrus are in relation with number of page views yet.

wizardlyhel:
I have 2 variation I am thinking to try:
- completely remove the use of `$.each` including the array case
- replace `$.each` logic to a jQuery one

What do you think @jansepar ?

jansepar:
+1 I think both those variations make sense to suss out.

donnielrt:
@Helen-Mobify we've pushed up a workaround for this for this particular project, right? 

I'm going to try and get some time allocated for next sprint so that we can investigate this a bit further.

wizardlyhel:
@jansepar is finding time to look at this again. I believe our action plan is to
1. Roll this PR out as a intermediate fix
2. Test out possible solutions to eliminate the last few drop thrus in additional PRs on top of the release

donnielrt:
@Helen-Mobify awesome! By rolling this PR, do you mean just building a bundle with this branch of mobifyjs? I don't think we should merge this in until we've fully tested it.

wizardlyhel:
I am not sure how we can automate a testing condition for this @donnielrt 
It happens so random and under odd conditions and is not reproducible locally.

and yes I do mean release this as a fix. (Possibly with some clean up as well)

wizardlyhel:
I tried the second fix variation where I replace `$.each` completely. This fix resulted in 0.005% fall thru rate, with medium at 0% over the course of 18 days for far.

donnielrt:
@Helen-Mobify do we know of any other clients affected by this bug?

wizardlyhel:
Anyone who are using mobifyjs is affected

donnielrt:
@Helen-Mobify FYI, we're creating a new [1.1.1 version of Mobify.js](https://github.com/mobify/mobifyjs/tree/v1.1.1) which includes your fixes, and not merging this fix directly into 1.1, so I'll go ahead and close this PR. 

This is to prevent an issue where a client with an older mobify-client doesn't miss out on this change (since their version of the 1.1 API will be different). Instead, they can now update their project to use the 1.1.1 in their `project.json`, and update their Mobify client.

PR #311, opened 2015-09-14
Apply scroll fix for ios8+ [1.0.x]

fractaltheory:
# ios8+ Scroll Fix

Status: **Ready for Review**

Reviewers: @donnielrt @haroldtreen  
JIRA: [RTM-335](https://mobify.atlassian.net/browse/RTM-335)
## Changes
- Apply scroll fix to ios8+
## How to Test
- Check out this branch
- `make`
- Switch a projects api to 1.0.1
- Test nothing breaks

### DIFF
--- a/api/main.js
+++ b/api/main.js
@@ -222,9 +222,9 @@ the browser behave as if the templated HTML was the regular source.
             Mobify.enhance();
         };
 
-        if (Mobify.isIOS8_0()) {
+        if (Mobify.isIOS8OrGreater(window.navigator.userAgent)) {
             // See `util.js` for more information about this fix.
-            Mobify.ios8_0ScrollFix(document, write);
+            Mobify.ios8AndGreaterScrollFix(document, write);
         } else {
             write();
         }
--- a/api/util.js
+++ b/api/util.js
@@ -81,14 +81,19 @@
         return i18nlookup;
     };
 
-    Mobify.isIOS8_0 = function() {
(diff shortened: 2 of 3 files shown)


jansepar:
Couple Qs:
1. Shouldn't this be a 1.0.2? Or are we just going to merge into 1.0.1 because noone is using 1.0.1 yet?
2. Why is this being merged into 1.0?

donnielrt:
Yeah, I think this should be a 1.0.2 release

fractaltheory:
There's no versioning or tags in use this far back - is there already a 1.0.1 that I would be clobbering?

Also, shouldn't it be merged into 1.0? That's the commit hash for the 1.0 submodule in mobify-client, isn't it: https://github.com/mobify/mobify-client/tree/master/vendor/mobify-js/1.0

donnielrt:
@fractaltheory sorry, I was thinking of 1.1.1. 1.0.1 should be fine.

@jansepar @noahadams and I were talking about this the other day, and the problem with merging new changes into 1.0 is that people who have an existing version of mobify-client installed will get a different 1.0 than people with a newer mobify-client. Namespacing it to 1.0.1 will mean that older clients will just explode when trying to preview/build.

@haroldtreen FYI ^^

Edit: clarified stuff

fractaltheory:
Closing in favor of leaving https://github.com/mobify/mobifyjs/tree/1.0 around as-is for posterity.

PR #318, opened 2019-05-29
Remove plaintext element for SEO

kevinxh:
We’ve had a number of projects have SEO issues, due to our capturing logic where a plaintext file is injected into our DOM. Googlebot will occasionally pick that up, and end up showing the contents of the plaintext file versus actual page meta data for the site.

This PR aims to remove the plaintext element as soon as we capture the document.

### DIFF
--- a/api/main.js
+++ b/api/main.js
@@ -36,6 +36,15 @@ $.extend(Mobify.transform, {
     // Read the conf, extract the Source DOM and begin the evaluation.
     prepareConf : function(rawConf) {
         var capturedState = Mobify.html.extractDOM();
+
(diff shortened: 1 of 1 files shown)


kevinxh:
Once this PR is merged, complete the TODOs in https://github.com/mobify/mobify-client/pull/88.

noahadams:
I would suggest adding a test, but I don't think there's an effective way to do that here now, looking at https://github.com/mobify/mobifyjs/blob/1.1.4/tests/index.html

noahadams (approved):
I buy that this is the right code in the right place YOLO :+1:

